PR #12, opened 2017-02-26
Rust setup

### LINKED ISSUE #9: Initial build setup for rust
If we want to port tools to rust, we obviously need to set up a build environment. A build should pull in [git2](https://crates.io/crates/git2) and produce a "git-dit"-library, which will at some point feature basic functionality used in the one-shot commands.
Blocks #10.

matthiasbeyer:
Closes #9 

---

This adds the setup for rewriting this in Rust. I created a "lib" crate, which can be used to contain the functionality, and a "bin" crate (which lives in the root of the project) which will be the "git-dit" executable, then.

This setup would result in _one_ binary for the whole `git-dit` project. I guess this is the desired way.

I would like to add another commit for the initial dependencies, but I want to have your opinion on this, first. I would go:

- [ ] libgit2 of course
- [ ] the normal "log" crate for logging
- [ ] clap for commandline parsing

What I would not include: configuration stuff, as we should use git for this. I guess `libgit2` has bindings for `git-config`, so that shouldn't be a problem, right?

---

We did not have a discussion whether we actually want to implement this in Rust, though. Do we? I would love to do this in Rust, and a extensive discussion might end in a lot of bikeshedding...


neithernut:
> This setup would result in one binary for the whole `git-dit` project. I guess this is the desired way.
Not exactly sure about that. I never especially liked bundling of tools in a single executable. One binary for one tool just feels right. Having a binary for each sub-command also makes porting easier since you can just keep the `git-dit` meta-command for now and swap out parts piece by piece. Git itself, for example, appears to have a binary for each tool in a `libexec`-directory.

I agree with you regarding the dependencies.

About using rust: well, we both already agreed that it'll be either C++ or rust.

neithernut:
> Not exactly sure about that. I never especially liked bundling of tools in a single executable. One binary for one tool just feels right. Having a binary for each sub-command also makes porting easier since you can just keep the git-dit meta-command for now and swap out parts piece by piece. Git itself, for example, appears to have a binary for each tool in a libexec-directory.

@matthiasbeyer and I were just discussing the matter in a private conversation. I do not object the single-binary approach. However, for the sake of extensibility we should search/execute a binary/script with a matching name if a sub-command is not implemented. This way, `git-dit` still mains extensible via 2rd-party scripts and programs.

PR #20, opened 2017-02-28
clap setup

matthiasbeyer:
This is my initial PR for porting the commandline to Rust.

This PR also introduces a `default.nix` file - if you think we shouldn't put these things in the repository, I'll happily remove it from my PR (and put an entry into `.gitignore` for it).

Please review the `cli.yml` carefully - there might be entries where the helptext/abouttext can be improved, and even more important: There might be entries where we can be more specific.

I did not yet extensively care about [the following configurations](https://docs.rs/clap/2.20.5/clap/struct.App.html):

* `before_help`
* `after_help`
* `requires`
* `conflicts_with`
* `settings`
* `group` / `groups`
* `display_order`

There is also no autocomplete-generating included in this PR.

### DIFF
--- a/src/cli.yaml
+++ b/src/cli.yaml
@@ -1,0 +1,378 @@
+name: git-dit
+version: 0.1.0
+authors:
+    - Julian Ganz <neither@nut.email>
+    - Matthias Beyer <mail@beyermatthias>
+about: Distributed issue tracker for git
+args:
+    - verbose:
+        short: v
+        long: verbose
+        multiple: false
+        help: Verbose output
+    - debug:
+        short: d
+        long: debug
+        multiple: false
+        help: Print debug output
+    - trace:
+        long: trace
+        multiple: false
+        help: Print trace output (very verbose debug logging)
+subcommands:
+    - check-message:
+        about: This command checks the message in file for validity
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - filename:
+                help: Path to file to check
+                index: 1
+                takes_value: true
+                multiple: false
+
+    - create-message:
+        about: >
+                 Create a new message. The parents provided will be the parents of the new
+                 message. If no issue hash is supplied, the new message will be the initial
+                 message of a new issue. Otherwise, the message will be associated with the
+                 issue hash.
+                 For the new message, the tree from the first parent supplied will be used.
+                 An appropriate reference will be created for the new message.
+                 Supplying an issue hash but no parent is considered an error.
+                 Returns (prints) the hash of the new commit.
+
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - parent-hash:
+                help: Hash of parent
+                index: 1
+                multiple: false
+            - tree-init-hash:
+                help: The hash of the initial commit in this issue tree
+                index: 2
+                multiple: false
+
+    - extract-trailers:
+        about: >
+                 Extract trailers from stdin and write them to stdout.
+                 This tool detects blocks of trailers and writes all trailers
+                 detected to stdout, in the order they were discovered. A block
+                 of trailers is a series of non-blank lines of text, with each
+                 single line being a trailer.
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+
+    - fetch:
+        about: Fetch issues
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - known:
+                help: also fetch known issues (those which visible through git dit list)
+                takes_value: false
+                multiple: false
+            - prune:
+                help: Prune (as with git fetch)
+                takes_value: false
+                multiple: false
+            - remote:
+                help: Remote to fetch from
+                index: 3
+                multiple: false
+            - issue:
+                help: Issue to fetch
+                index: 4
+                multiple: true
+
+    - find-tree-init-hash:
+        about: This command prints the init hash of the issue commit belongs to.
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - commit:
+                help: commit hash
+                index: 1
+                multiple: false
+
+    - get-issue-metadata:
+        about: >
+                 Prints a log of commit tags, from the supplied issue head to
+                 the initial issue message.
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - head:
+                help: The head for which to collect the metadata
+                index: 1
+                multiple: false
+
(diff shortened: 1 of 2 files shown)


src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+authors:
+    - Julian Ganz <neither@nut.email>
+    - Matthias Beyer <mail@beyermatthias>
+about: Distributed issue tracker for git
+args:
+    - verbose:
+        short: v
+        long: verbose
+        multiple: false
+        help: Print what you are doing
```
neithernut: This doesn't seem right. Verbose output doesn't just contain what the program does. "Verbose output" or something similarly vague would be fine.

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+    - create-message:
+        about: >
+                 If <parent-hash> and <tree-init-hash> are not provided, we
+                 create a new issue.
+                 If <parent-hash> is provided and <tree-init-hash> is not, we
+                 create a new issue which is attached to an existing commit.
+                 That can be used (for example) to attach a bug report to a
+                 commit.
+                 If both are provided, we create a reply to an existing issue
+                 discussion.
```
neithernut: I presume you copied this from the comment at the top of the script. Please use the help text instead. The functionality has changes somewhat. (I apparently forgot to update the comment)

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+            - parent-hash:
+                help: Hash of parent
+                index: 1
+                multiple: false
+            - tree-init-hash:
+                help: The hash of the initial commit in this issue tree
+                index: 2
+                multiple: false
+
+    - extract-trailers:
```
neithernut: Not sure whether we really want to port this tool. Internally, we will most likely use another representation of tags which we'll parse from commit messages. On the other hand, it _may_ enhance scriptability.

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+
+    - fetch:
+        about: Fetch issues
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - known:
+                help: also fetch known issues (this which visible through git dit list
```
neithernut: A parentheses is missing. The contents of the parentheses seems to be not right, either.

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+
+    - find-tree-init-hash:
+        about: This command prints the init hash of the issue a commit belongs to.
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - parent:
+                help: parent hash
```
neithernut: Parent hash? We take in the hash of the commit itself, don't we?

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                short: M
+                long: metadata
+                help: >
+                        Add metadata. Can be passed multiple times.
+                        Metadata is added and removed via key-value pairs.
+                        Available Keys are:
+                            status=<value>
+                            priority=<value>
+                            severity=<value>
+                            assignee=<value>
```
neithernut: Now that we have long options, we may as well ass long options for each of those. The current mechanism is not especially robust (considering typos).

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                        where <value> is the desired new value to set.
+                        Passing an empty value leaves the metadata empty.
+                multiple: true
+                takes_value: true
+                value_names:
+                    - data
+
+
+    - prepare-metadata:
```
neithernut: We don't need this anymore.

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                multiple: false
+                takes_value: false
+            - gpgsign:
+                short: S
+                long: gpg-sign
+                help: Add a GPG signature
+                multiple: false
+                takes_value: false
+            - tempfile:
+                short: t
```
neithernut: Not sure whether we should really have a short option for this, anymore.

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                value_names:
+                    - path
+            - quote:
+                short: q
+                long: quote
+                help: Quote parent
+                multiple: false
+                takes_value: false
+            - reference:
+                short: e
```
neithernut: Wasn't this supposed to be an `r`?

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                    - commithash
+
+    - show:
+        about: This uses 'git log' to print the issues.
+        version: 0.1.0
+        authors:
+            - Matthias Beyer <mail@beyermatthias.de>
+            - Julian Ganz <neither@nut.email>
+        args:
+            - parent:
```
neithernut: What's this for?

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                conflicts_with:
+                    - set-status
+            - set-status:
+                short: s
+                long: status
+                help: Set a status tag. Key-value pair expected.
+                multiple: false
+                takes_value: false
+            - reference:
+                short: e
```
neithernut: This should also be an `r`.

src/cli.yaml
```diff
@@ -0,0 +1,395 @@
...
+                long: list
+                help: List Tags. Lists all tags which were introduced in the discussion thread of the issue
+                multiple: false
+                takes_value: false
+                conflicts_with:
+                    - set-status
+            - set-status:
+                short: s
+                long: status
+                help: Set a status tag. Key-value pair expected.
```
neithernut: If we call it "status" it should just set `Git-status`. For other metadata, the interface of `reply` should be mirrored.

src/cli.yaml
matthiasbeyer: I just took the comment from the script here.

src/cli.yaml
matthiasbeyer: Yes, but we would need another generical flag for specifying arbitrary content, don't we?

matthiasbeyer: I could introduce a new subcommand for this, if you want.

src/cli.yaml
matthiasbeyer: I copied this from the script documentation... :smile:

neithernut: Well, change this to `message`. I suspect that our initial use case for this subcommand somehow slipped into the documentation.

src/cli.yaml
neithernut: We should definitely have long-options for the "standard"-tags.

We could still have an option for adding general-purpose tags (not only ones starting with `Dit-`). This option should, however, not refer to the keys we actually may use within `git-dit`. This would encourage using this option despite existing sane options. Also, it should not be a short option, imo.

A subcommand is not necessary.

src/cli.yaml
neithernut: Mind renaming this parameter?

neithernut:
Blocks #23.

src/cli.yaml
matthiasbeyer: Renamed it to `"commit"`, which I find more appropriate (also because this is plumbing).

src/cli.yaml
matthiasbeyer: Well, I'd let that specific part to you for a follow-up PR, actually, as I'm not sure what my opinion is on this one.

matthiasbeyer:
Ready to squash here, as soon as you approve.

matthiasbeyer:
Squashed.

PR #22, opened 2017-02-28
Add repository module

matthiasbeyer:
This is the tracking-PR for the library implementation.

I would model this as extension to the `git2::Repository` type. Further, I would model each shell script into one function on `Repository`.

I would like to hear your opinion before digging further into this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,9 @@ which is, as of 2017-02-24, git version `2.11.1`.
 
 ---
 
-Licensed under the terms of GNU GPLv2.
-For more information, see the LICENSE file.
+We are rewriting this application.
+The [library module](./lib) is licensed under terms of [MPL-2.0](./lib/LICENSE).
+The binary module (this directory) uses the library and provides a commandline
+interface for it and is licensed under terms of [GNU GPLv2](./LICENSE).
 
 (c) Julian Ganz, Matthias Beyer
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -6,6 +6,10 @@ authors = ["Matthias Beyer <mail@beyermatthias.de>",
 
 workspace = "./.."
 
+repository    = "https://github.com/neithernut/git-dit/tree/master/lib"
+readme        = "./README.md"
+license       = "MPL-2.0"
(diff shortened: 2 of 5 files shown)


matthiasbeyer:
Ah, and another thing: Do we want this to be GPL? I would say no, as this is a library others might want to use. As far as I know, LGPL does not work with Rust, as things are statically linked - so the best option would be MPL IMHO.

Profpatsch:
> I would say no, as this is a library others might want to use.

Isn’t that what GPL is for?

neithernut:
GPL prevents using a library in binaries licensed under more permissive licenses, like MIT and BSD. With rust, this issue is even more present, as the linking exception granted e.g. by LGPL doesn't apply, as there is not really a difference between implementation and interface (regarding the build process).

Profpatsch:
> GPL prevents using a library in binaries licensed under more permissive licenses, like MIT and BSD.

Which is a feature imo. Of course, that’s for you to decide.

neithernut:
> I would model this as extension to the `git2::Repository` type.

Yep. I'd have done the same thing.

> Further, I would model each shell script into one function on Repository.

With C++, I'd prefer private free functions. But with rust this seems to be a reasonable approach.

----

> > GPL prevents using a library in binaries licensed under more permissive licenses, like MIT and BSD.
>
> Which is a feature imo. Of course, that’s for you to decide.

It's an anti-feature in my opinion. Libraries are there to be used, especially those providing infrastructure. I would rather have devs working on non-free products use a library that's open source than have them build hafl-assed, broken crap.

If they don't want to contribute, have their own, private development and one day realize they have lost the possibility to incorporate upstream fixes and features in their products, that's not my problem.

matthiasbeyer:
And this is wäre MPL steps in, afaik. With MPL you can use the library as you please, butbif you change it you have to contribute back.

So it's LGPL for Rust, essentially.

Either way, obligatorily: IANAL!

neithernut:
I assuming that you are referring to the Mozilla Public License (not the Microsoft Public License). I would accept releasing the library under MPL2.0.

Issue #10, opened 2017-02-26
Port find-tree-init-hash to rust


neithernut:
Add a function implementing the search. Then add a program calling that function, replacing the existing sub-command.
Blocks #11.

PR #23, opened 2017-03-04
Add preliminary implementation of find_tree_init_hash

### LINKED ISSUE #24: Add iterator for first-parent-iteration of commits
For some of the functionality we want to provide we need to trace commits back to an issue's initial message via the chain of commits' first-parents. An iterator which follows the first parents would be advantageous.
The iterator should be constructible from a `git2::Commit`, return the commit itself on the first call to `next()` and return the first-parent of the last commit returned in subsequent calls.

neithernut:
Targets #10. Resolves #24.

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -12,3 +12,5 @@ extern crate git2;
 
 pub mod repository;
 
+mod first_parent_iter;
+
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -11,6 +11,7 @@
 #[macro_use] extern crate error_chain;
 extern crate git2;
 
+pub mod error;
 pub mod repository;
 
 mod first_parent_iter;
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -8,6 +8,7 @@
 //
 
 #[macro_use] extern crate log;
+#[macro_use] extern crate error_chain;
 extern crate git2;
 
 pub mod repository;
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,27 @@
+// git-dit - the distributed issue tracker for git
+// Copyright (C) 2017 Matthias Beyer <mail@beyermatthias.de>
+// Copyright (C) 2017 Julian Ganz <neither@nut.email>
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 2 as
+// published by the Free Software Foundation.
+//
+
+error_chain! {
+    foreign_links {
+        GitError(::git2::Error);
+        GitDitError(::libgitdit::error::Error);
+    }
+
+    errors {
+        WrappedGitError {
+            description("TODO: Wrapped error")
+            display("TODO: Wrapped error")
+        }
+
+        WrappedGitDitError {
+            description("TODO: Wrapped error")
+            display("TODO: Wrapped error")
+        }
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,6 +17,7 @@ mod error;
 
 use clap::App;
 use git2::Repository;
+use libgitdit::repository::RepositoryExt;
 
 use error::ErrorKind as EK;
 use error::*;
@@ -30,6 +31,19 @@ use error::*;
 fn open_dit_repo() -> Result<Repository> {
     // TODO: access the config and maybe return another repo instead
     Repository::open_from_env().chain_err(|| EK::WrappedGitError)
+}
+
+
+/// find-tree-init-hash subcommand implementation
+///
+fn find_tree_init_hash(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
+    // note: commit is always present since it is a required parameter
+    repo.revparse_single(matches.value_of("commit").unwrap())
+        .and_then(|obj| repo.find_commit(obj.id()))
+        .chain_err(|| EK::WrappedGitError)
+        .and_then(|commit| repo.find_tree_init(commit).chain_err(|| EK::WrappedGitDitError))
+        .map(|commit| {println!("{}", commit.id()); 0})
+        .unwrap_or_else(|err| {error!("{}", err); 1})
 }
 
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,22 @@ extern crate libgitdit;
 mod error;
 
 use clap::App;
+use git2::Repository;
+
+use error::ErrorKind as EK;
+use error::*;
+
+
+/// Open the DIT repo
+///
+/// Opens the DIT repo corresponding to the current one honouring the user
+/// configuration.
+///
+fn open_dit_repo() -> Result<Repository> {
+    // TODO: access the config and maybe return another repo instead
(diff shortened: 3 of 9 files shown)


neithernut:
For now, I'm reusing the `git2::Error` type. I guess @matthiasbeyer would like to use a new one for `git-dit`.

matthiasbeyer:
Yep. I'd add `error_chain` to the dependencies for this.

neithernut:
Do you want to open a PR for introducing an error type or should I do it in this one?

matthiasbeyer:
I'm about to open one. Hold on.

neithernut:
I will now rebase to get the clap setup.

src/error.rs
```diff
@@ -0,0 +1,32 @@
...
+            description("TODO: Wrapped error")
+            display("TODO: Wrapped error")
+        }
+
+        SubCommandNotFound(subcommand: String) {
+            description("Unknown subcommand")
+            display("Unknown subcommand: {}", subcommand)
+        }
+    }
+}
```
neithernut: This file was largely copied from `lib/src/error.rs`, which is licensed under the MPL. Are you ok with this, @matthiasbeyer?

neithernut:
It doesn't work yet. It also lacks a check for the presence of the `commit` parameter.

neithernut:
I also just realized that clap ships with an error type

src/main.rs
```diff
@@ -9,12 +9,51 @@
...
+fn open_dit_repo() -> Result<Repository> {
+    // TODO: access the config and maybe return another repo instead
+    Repository::open_from_env().chain_err(|| EK::WrappedGitError)
+}
+
+
+/// find-tree-init-hash subcommand implementation
+///
+fn find_tree_init_hash(repo: &Repository, matches: &clap::ArgMatches) -> Result<()> {
+    repo.refname_to_id(matches.value_of("commit").unwrap())
```
neithernut: Apparently, `refname_to_id()` doesn't do what I thought it does.

neithernut:
Works now, apparently. Stll, I'm not sure whether I should discard the `FirstParentIter` in favour for [git2::Revwalk](https://docs.rs/git2/0.6.4/git2/struct.Revwalk.html) or if our use-case justifies the extra iterator (for convenience).

src/error.rs
matthiasbeyer: Yes.

neithernut:
Blocks #28.

src/main.rs
```diff
@@ -9,12 +9,51 @@
...
+fn open_dit_repo() -> Result<Repository> {
+    // TODO: access the config and maybe return another repo instead
+    Repository::open_from_env().chain_err(|| EK::WrappedGitError)
+}
+
+
+/// find-tree-init-hash subcommand implementation
+///
+fn find_tree_init_hash(repo: &Repository, matches: &clap::ArgMatches) -> Result<()> {
+    repo.revparse_single(matches.value_of("commit").unwrap())
```
matthiasbeyer: Please add a short comment here that `unwrap()` is secured by clap.

src/main.rs
```diff
@@ -9,12 +9,51 @@
...
     let matches = App::from_yaml(yaml).get_matches();
-    println!("Hello, world!");
+
+    let repo = open_dit_repo().unwrap();
+
+    match matches.subcommand() {
+        ("find-tree-init-hash", Some(sub_matches))  => find_tree_init_hash(&repo, sub_matches),
+        (name, _) => Err(Error::from_kind(EK::SubCommandNotFound(String::from(name)))),
+        // TODO: find and exec 3rd party executable
+    }.unwrap()
```
matthiasbeyer: This panics on error (for example if `find_tree_init_hash()` returns an `Err(_)`.

What I'd do: Return nothing from `find_tree_init_hash()` and let the function itself do the printing. We could return an exit value from the function,... which then gets used in main() to `std::process::exit()`...

matthiasbeyer:
> Works now, apparently. Stll, I'm not sure whether I should discard the FirstParentIter in favour for git2::Revwalk or if our use-case justifies the extra iterator (for convenience).

I'd use `git2::Revwalk` and wrap it in our iterator. Overhead is almost zero then, but we get the flexibility and we're able to tweak it later on without rewriting the codebase.

src/main.rs
neithernut: Yep. Originally, I planned on implementing proper error handling later on. But I'll implement the changes in this PR. This will also simplify implementing #28.

src/main.rs
```diff
@@ -36,24 +36,26 @@ fn open_dit_repo() -> Result<Repository> {
...
+        .unwrap_or_else(|err| {error!("{}", err); 1})
 }
 
 
 fn main() {
     let yaml    = load_yaml!("cli.yaml");
     let matches = App::from_yaml(yaml).get_matches();
 
-    let repo = open_dit_repo().unwrap();
+    let repo = open_dit_repo().unwrap(); // TODO: error handling
```
matthiasbeyer: `TODO: Fix panic` would be way more accurate :smile: 

Also, we can do a simple 

```
let repo = match open_dir_repo() { 
    Ok(r) => r, 
    Err(e) => { 
        error!("{}", e); 
        std::process::exit(1) 
    }
}
```

 here

src/main.rs
```diff
@@ -36,24 +36,26 @@ fn open_dit_repo() -> Result<Repository> {
...
 
-    match matches.subcommand() {
+    std::process::exit(match matches.subcommand() {
         ("find-tree-init-hash", Some(sub_matches))  => find_tree_init_hash(&repo, sub_matches),
-        (name, _) => Err(Error::from_kind(EK::SubCommandNotFound(String::from(name)))),
```
matthiasbeyer: Why did you remove this?

src/main.rs
neithernut: Well, yeah. I was kinda lazy.

src/main.rs
neithernut: I did not remote this but replace it with direct error printing.

matthiasbeyer: Whops. sure! :laughing:

neithernut:
About the `FirstParentIter`: apparently `Revwalk` must be pulled from a repo while our iterator can be created from a commit. Hence, I'd keep the current implementation.

src/main.rs
neithernut: I also did not realize that this doesn't generate any conflicts between the branches' return type.

neithernut:
I'm ready for squashing.

matthiasbeyer:
Go ahead.

neithernut:
@matthiasbeyer can you give me an ACK/NACK on this?

lib/src/lib.rs
```diff
@@ -8,7 +8,11 @@
 //
 
 #[macro_use] extern crate log;
+#[macro_use] extern crate error_chain;
 extern crate git2;
 
+pub mod error;
 pub mod repository;
 
+mod first_parent_iter;
```
matthiasbeyer: I'm rather sure we need to make this `pub` or export the iterator in another way... But I guess we can do so later.

Issue #19, opened 2017-02-27
Add sub-command calling mechanism in rust port


neithernut:
We need some infrastructure for calling the sub-command internally. If a sub-command is not implemented in the binary, if should try to invoke a binary with a matching name.
Blocks #10 and #11.

PR #28, opened 2017-03-04
Add mechanism for invokation of 3rd party executables

neithernut:
Targets #19.


neithernut:
The branch will be rebased at some point.

src/main.rs
```diff
@@ -9,12 +9,73 @@
...
+
+
+/// Handle unknown subcommands
+///
+/// Try to invoke an executable matching the name of the subcommand.
+///
+fn handle_unknown_subcommand(name: &str, matches: &clap::ArgMatches) -> Result<()> {
+    let mut command = Command::new(format!("git-dit-{}", name));
+    let result = try!(matches.values_of("")
+                             .unwrap()
```
neithernut: Currently, this will make the entire thing fail if the subcommand is invoked with zero arguments.

PR #31, opened 2017-03-06
Port get-issue-tree-init-hashes

### LINKED ISSUE #30: Port get-issue-tree-init-hashes to rust
Add a function or function family to our `RepositoryExt` returning issue hashes (e.g. as an iterator of `git2::Oid`s). It should be possible to retrieve the hashes for a given set of remotes and/or the local repository.
Then add an implementation of the `get-issue-tree-init-hashes` subcommand to the binary.

neithernut:
Resolves #30.

Currently, it isn't ensured that a hash isn't printed twice, e.g. if two removes have a head-reference for an issue.


lib/src/repository.rs
```diff
@@ -27,8 +27,40 @@ pub trait RepositoryExt {
...
+}
+
+
+/// Get issue hashes from head references
+///
+/// Retrieve the issue hashes from head references provided. References which
+/// are not head references are ignores. However, the function does not check
+/// whether the references are, in fact, references in a dit namespace.
+///
+fn _head_refs_to_issues(references: References) -> Vec<Oid> {
```
matthiasbeyer: No need for a leading `_` underscore!

lib/src/repository.rs
```diff
@@ -27,8 +27,40 @@ pub trait RepositoryExt {
...
+    references.names()
+              .filter_map(|name| name.ok())
```
matthiasbeyer: Hiding errors here, a `// TODO` would be nice.

lib/src/repository.rs
```diff
@@ -27,8 +27,40 @@ pub trait RepositoryExt {
...
+    /// prefix provided (e.g. all issues for which refs exist under
+    /// `<prefix>/dit/`). Provide "refs" as the prefix to get only local issues.
+    ///
+    fn get_issue_hashes(&self, prefix: &str) -> Result<Vec<Oid>>;
+
+    /// Get all issue hashes
+    ///
+    /// This function returns all known issues known to the DIT repo.
+    ///
+    fn get_all_issue_hashes(&self) -> Result<Vec<Oid>>;
```
matthiasbeyer: Returning an iterator would be way better IMO, and the current implementation would allow that easily.

neithernut: An unsized trait cannot be wrapped in a result or returned directly and I didn't feel it would be the right way to explicitly write out the type. Or does Rust have something similar to the `auto` keyword on C++?

lib/src/repository.rs
neithernut: I felt it was ok to simply drop unnamed references, as they cannot be issue heads. However, I can check for possibilities to propagate the error instead.

neithernut: Forwarding an error here is pretty cumbersome. For such things, we would want an iterator extension which allows aborting the iteration in some way.

lib/src/repository.rs
matthiasbeyer: What are you talking about? Returning an iterator is ` -> Result<IssueHashIterator>` should work perfectly well...

neithernut: I obviously would have to declare that type.

matthiasbeyer: Yes.

neithernut: Well, the type declaration will either be utterly complicated (due to all the maps involved), or consist of a struct wrapping a generic iterator. I would go with the second one.

matthiasbeyer: Definitively the second one.

lib/src/repository.rs
```diff
@@ -7,13 +7,17 @@
...
-
 use git2::{Commit, Oid, References, Repository};
+
+use boxed_iterator::BoxedIterator;
+use error::*;
+use error::ErrorKind as EK;
 use first_parent_iter::FirstParentIter;
 
 
+pub type HashIterator<'a> = BoxedIterator<'a, Oid>;
```
matthiasbeyer: I would rather call this an `OidIterator`.

lib/src/repository.rs
```diff
@@ -27,8 +31,40 @@ pub trait RepositoryExt {
...
+/// whether the references are, in fact, references in a dit namespace.
+///
+fn head_refs_to_issues(refs: References) -> HashIterator {
+    let hashes = refs.names()
+                     .filter_map(|name| name.ok()) // TODO: propagate errors
+                     .filter(|name| name.ends_with("/head"))
+                     .filter_map(|name| name.rsplitn(3, "/")
+                                            .nth(1)
+                                            .and_then(|hash| Oid::from_str(hash).ok()));
+    HashIterator::new(hashes)
```
matthiasbeyer: If you implemennt `From<I> for BoxedIterator<I> where I: Iterator<...>` we could reduce this boilerplate and implement this function as:

```rust
refs.names()
     .filter_map(|name| name.ok()) // TODO: propagate errors
     .filter(|name| name.ends_with("/head"))
     .filter_map(|name| name.rsplitn(3, "/").nth(1).and_then(|hash| Oid::from_str(hash).ok()));
```

I guess.

lib/src/boxed_iterator.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next()
+    }
+}
+
+impl<'a, T> FromIterator<T> for BoxedIterator<'a, T> {
+    // TODO: the lifetime 'a in the declaration of `from_iter()` imposes
+    //       stricter restrictions on `I` than the trait.
+    fn from_iter<I: IntoIterator<Item=T> + 'a>(iterator: I) -> Self {
```
neithernut: @matthiasbeyer any ideas on this one?

neithernut:
I rolled back most of the changes. @matthiasbeyer will introduce an iterator for the specific use-case. I really don't like this, but it looks like it's the only option for now. If rust allows returning traits at some point in the future, we may just do that instead.

src/main.rs
```diff
@@ -48,6 +48,24 @@ fn find_tree_init_hash(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
 
+/// find-tree-init-hash subcommand implementation
+///
+fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
+    match repo.get_all_issue_hashes() {
+        Ok(hashes)  => {
+            for hash in hashes {
+                match hash {
+                    Ok(hash) => println!("{}", hash),
+                    Err(e)   => error!("{:?}", e),
```
matthiasbeyer: I guess we need to `return 1` here, do we?

lib/src/iter.rs
```diff
@@ -0,0 +1,48 @@
...
+            })
+    }
+}
+
+impl<'r> From<References<'r>> for HeadRefsToIssuesIter<'r> {
+    fn from(r: References<'r>) -> HeadRefsToIssuesIter<'r> {
+        HeadRefsToIssuesIter(r.names())
+    }
+}
+
```
matthiasbeyer: We could `impl From<RefrerenceNames> for HeadRefsToIssuesIter` here as well.

PR #34, opened 2017-03-08
Spawn editor helper

### LINKED ISSUE #33: Add a utility function for spawning an editor
The `new` and `reply` subcommands will spawn an editor. We should add a function doing this, honouring the git-config. The function should resort to the `EDITOR` variable, if the config doesn't specify an editor.

This partially targets #17, albeit the issue will only be resolved when both `new` and `reply` are ported to rust.

matthiasbeyer:
Closes #33


lib/src/error.rs
```diff
@@ -20,6 +20,16 @@ error_chain! {
...
+        ConfigError(configname: String) {
+            description("Could not find config")
+            display("Could not find config: {}", configname)
+        }
+
+        EditorError {
+            description("Could not find editor configuration or ENV variable")
+            display("Could not find editor configuration or ENV variable")
+        }
+
```
neithernut: This snippet should also be moved to the binary

matthiasbeyer: whops.

PR #35, opened 2017-03-08
Trailers

### LINKED ISSUE #27: Port extract-trailers to rust
Device on a representation of trailers (as key-value pairs). Simple tuples might suffice.
Add a facility for extracting them from a stream (e.g. iterator of chars). This may be a function to `RepositoryExt` or (more fancy) a iterator-adapter. Implement `extract-trailers` based on the function.

matthiasbeyer:
~~Closes~~ Targets #27!
Very WIP!
Oh so Rust!
Much Iterator!
Have Review!
Many Nice!


lib/src/trailer.rs
```diff
@@ -0,0 +1,116 @@
...
+        loop {
+            match self.0.next() {
+                None => return None,
+                Some(trailer) => {
+                    if trailer.key.0.starts_with("Dit") {
+                        return Some(trailer);
+                    } else {
+                        continue;
+                    }
+
```
matthiasbeyer: Blank line to be removed.

lib/src/trailer.rs
```diff
@@ -0,0 +1,116 @@
...
+                    match RE.captures(line) {
+                        None => continue,
+                        Some(capture) => {
+                            let key = capture.name("key").map(|m| {
+                                TrailerKey(String::from(m.as_str()))
+                            }).unwrap(); // TODO: fix unwrap()
+
+                            let value = capture.name("value").map(|m| {
+                                TrailerValue::from_slice(m.as_str())
+                            }).unwrap(); // TODO: fix unwrap()
```
matthiasbeyer: Must clearly be fixed here, so I block my own PR with this.

lib/src/trailer.rs
```diff
@@ -0,0 +1,116 @@
...
+        loop {
+            match self.0.next() {
+                None => return None,
+                Some(line) => {
+                    match RE.captures(line) {
+                        None => continue,
+                        Some(capture) => {
+                            let key = capture.name("key").map(|m| {
+                                TrailerKey(String::from(m.as_str()))
+                            }).unwrap(); // TODO: fix unwrap()
```
matthiasbeyer: Must clearly be fixed here, so I block my own PR with this.

lib/src/trailer.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+}
+
+impl<'a> Iterator for Trailers<'a> {
+    type Item = Trailer;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"^(?P<key>(.*)):\ (?P<value>(.*))$").unwrap();
```
matthiasbeyer: I did not verify whether this is the correct regex.

lib/src/trailer.rs
```diff
@@ -40,6 +40,12 @@ pub struct Trailer {
     pub value: TrailerValue,
 }
 
+impl Into<(TrailerKey, TrailerValue)> for Trailer {
```
matthiasbeyer: I guess this is automatically implemented with the `From` implementation. @TheNeikos?

TheNeikos: Yes

neithernut:
In case you need some inspiration regarding the points i lined out in my review:

If I were to make this iterator capable of ignoring text paragraphs, I'd introduce an enum
```
LineContent {
    Trailer(Trailer),
    Text(String),
    Blank,
}
```
and an additional `Iterator<LineContent>`. This iterator would categorize lines of text into (multiline) trailers, regular lines of text and empty/blank lines.

`Trailers` would have an internal buffer for `Trailer` structs, from which it would return one on each call to `next()`. If the buffer is empty, it would use the aforementioned iterator for populating the buffer, discarding the contents for an entire block if a line of text was encountered.

neithernut:
Additional commits in matthiasbeyer/git-dit#1.

matthiasbeyer:
Merged your commits, rebased and squashed a few commits of the whole PR. Ready for your review now.

lib/src/message/line.rs
```diff
@@ -0,0 +1,85 @@
...
+pub enum Line {
+    Text(String),
+    Trailer(Trailer),
+    Blank
+}
+
+impl<'a> From<&'a str> for Line {
+    fn from(line :&'a str) -> Self {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"^(?P<key>([^[:space:]]+)):\ (?P<value>(.*))$").unwrap();
```
TheNeikos: Maybe change RE to a variable that describes what you are trying to do

lib/src/message/trailer.rs
```diff
@@ -0,0 +1,211 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+
+use message::line::{Line, Lines};
+use std::collections::LinkedList;
+
+#[derive(Debug, Hash, Eq, PartialEq, Ord, PartialOrd, Clone)]
+pub struct TrailerKey(String);
```
TheNeikos: Why not a `type TrailerKey = String`

lib/src/message/trailer.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+}
+
+use std::collections::BTreeMap;
+use std::collections::HashMap;
+
+pub trait IntoMap {
+    fn into_hashmap(self) -> HashMap<TrailerKey, TrailerValue>;
```
TheNeikos: You don't need this.

You can `collect` into `HashMap<K, V>` if your iterator returns `(K, V)`. 

https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.collect

lib/src/message/trailer.rs
```diff
@@ -0,0 +1,211 @@
...
+            },
+            TrailerCollector::Dumping => self,
+        }
+    }
+}
+
+
+pub struct Trailers<'a> {
+    lines: Lines<'a>,
+    buf: LinkedList<Trailer>,
```
TheNeikos: Why a linked list? Why not just a Vector?

lib/src/message/trailer.rs
matthiasbeyer: Because I had bad experiences with newtypes (you remember the `StoreId` desaster with imag?)!

lib/src/message/trailer.rs
matthiasbeyer: I guess @neithernut would argue that a LL is faster. It doesn't really matter, though.

neithernut: Because the ordering will be inversed with a vector.

TheNeikos: If you really need `pop_front` you can use: `VecDeque`

matthiasbeyer: TIL.

lib/src/message/trailer.rs
TheNeikos: To be more specific, this part of the implementation is here: https://doc.rust-lang.org/stable/std/iter/trait.FromIterator.html

lib/src/message/trailer.rs
matthiasbeyer: @neithernut care to change this (and push onto this very PR)?

lib/src/message/trailer.rs
TheNeikos: Alright. Also, just to make sure the correct terminology is used.

Newtype => `struct Foo(String)`
Type Alias => `type Foo = String`

lib/src/message/trailer.rs
neithernut: Yep.

lib/src/message/trailer.rs
matthiasbeyer: Okay, got the wording wrong. You know what I mean.

lib/src/message/trailer.rs
```diff
@@ -0,0 +1,196 @@
...
+        'outer: loop {
+            if let Some(trailer) = self.buf.pop_front() {
+                return Some(trailer);
+            }
+
+            // refill buffer from next block
+            let mut collector = TrailerCollector::new(&mut self.buf);
+            let mut at_end = true;
+
+            'refill: for line in self.lines.next() {
```
TheNeikos: This is not used

PR #39, opened 2017-03-09
Add a macro for early returns in subcommands

neithernut:
Add a macro which hopefully safes us a few lines of code. I'm happy to change the name if someone comes up with a better alternative.


neithernut:
Blocks #40.

PR #40, opened 2017-03-09
Port "create-message" to rust

### LINKED ISSUE #11: Port create-message to rust
What the title says. For this, a function may be introduced to the library for checking the message format.

neithernut:
Resolves #11.


lib/src/repository.rs
```diff
@@ -41,6 +41,28 @@ pub trait RepositoryExt {
     /// This function returns all known issues known to the DIT repo.
     ///
     fn get_all_issue_hashes(&self) -> Result<HeadRefsToIssuesIter>;
+
+    /// Create a new message
+    ///
+    /// This function creates a new issue message as well as an appropriate
+    /// reference and return the new issue's hash. If the issue supplied is
+    /// none, the message will be the initial message of a new issue and the
+    /// hash returned will be also the hash of the new issue.
```
matthiasbeyer: Sorry, but this is not clear for me. Can your reword the documentation for this function, please?

src/main.rs
```diff
@@ -75,6 +75,29 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+                       .and_then(|i| Oid::from_str(i).ok()); //TODO: this hides errors
+    let sig = try_or_1!(repo.signature());
+
+    let parents : Vec<Commit> = match matches.values_of("parents") {
+        Some(values) => values.filter_map(|parent| Oid::from_str(parent).ok()) //TODO: this hides errors
+                              .filter_map(|oid| repo.find_commit(oid).ok()) //TODO: this hides errors
+                              .collect(),
+        _            => Vec::new(),
+    };
+    let parent_refs : Vec<&Commit> = parents.iter().map(|command| command).collect();
```
matthiasbeyer: Why not merging this piece of code with the above `collect()`? I know we want `&[&Commit]` instead of `&[Commit]`, but it should be possible without this boilerplate line!

src/main.rs
```diff
@@ -75,6 +75,29 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+
+    // TODO: use the first parent's tree if availible, as advertised in the documentation
+    let tree = try_or_1!(repo.empty_tree());
+
+    try_or_1!(repo.create_message(issue.as_ref(), &sig, &sig, "", &tree, &parent_refs));
```
matthiasbeyer: Why do you call it with an empty message here?

neithernut: Because it's WIP.

src/main.rs
neithernut: Nope, it doesn't work because the commits don't live long enough if they aren't stored, e.g. in a `Vec<Commit>`.

neithernut:
Some issues surfaced during testing.

neithernut:
The issues I observed were resolved.

src/main.rs
```diff
@@ -52,6 +52,23 @@ fn open_dit_repo() -> Result<Repository> {
 }
 
 
+/// Get a vector of commits from values
+///
+/// This function transforms values to a vector. If no values are supplied,
+/// an empty vector will be returned.
+///
+fn values_to_hashes<'repo>(repo: &'repo Repository, values: Option<Values>) -> Result<Vec<Commit<'repo>>> {
```
matthiasbeyer: The function is all about the `Values`, but the parameter for that is an `Option` - not very clean IMO.

neithernut:
Looks like I broke things with a rebase.

src/main.rs
neithernut: It was intended to reduce boiler plate. For some reason the clap guys chose to return `None` instead of just an empty `Values` in cases where the option was not supplied.

lib/src/repository.rs
```diff
@@ -75,5 +98,34 @@ impl RepositoryExt for Repository {
...
+        // commit message
+        let msg_id = try!(self.commit(None, author, committer, message, tree, parents));
+
+        // make an apropriate reference
+        let refname =  match issue {
+            Some(hash)  => format!("refs/dit/{}/leaves/{}", hash, msg_id),
+            _           => format!("refs/dit/{}/head", msg_id),
+        };
+        let logmsg = format!("new message: {}", msg_id);
+        try!(self.reference(&refname, msg_id, false, &logmsg));
```
matthiasbeyer: A comment that the `logmsg` is for the reflog would be ... appropriate I think.

src/main.rs
```diff
@@ -52,6 +57,20 @@ fn open_dit_repo() -> Result<Repository> {
...
+/// This function transforms values to a vector.
+///
+fn values_to_hashes<'repo>(repo: &'repo Repository, values: Values) -> Result<Vec<Commit<'repo>>> {
+    let mut retval = Vec::new();
+    for commit in values.map(|string| repo.revparse_single(string))
+                        .map(|oid| repo.find_commit(try!(oid).id())) {
+        retval.push(try!(commit));
+    }
+    Ok(retval)
+}
```
matthiasbeyer: It is not really nice that you collect the iterator here and start a new iterator from the `Vec` below, but I let that slip because I don't see any value in optimizing this right now and we can improve it later anyways.

matthiasbeyer:
Please read my comments and don't forget to rebase! :smile:

src/main.rs
neithernut: Not collecting isn't an option:

 * We need to know in advance whether an error occurred or not.
 * The commits have to be stored somewhere anyways.

PR #43, opened 2017-03-10
patches for: Add create message

matthiasbeyer:
One patch for the "Add create message" PR #40


matthiasbeyer:
Forget what I said... this is BS.

neithernut:
Yep. There's a reason I didn't use closures. But maybe I should add some comments about that somewhere.

PR #48, opened 2017-03-17
Add check_message

### LINKED ISSUE #26: Port check-message to rust
What the title says.

neithernut:
Resolves #26.


lib/src/message/mod.rs
```diff
@@ -17,13 +17,13 @@ pub mod trailer;
...
-/// line, followed by an empty line.
+/// line, followed by an empty line. The message is supplied as an iterator over
+/// the lines of the message.
 ///
 /// Note that the function expects a message which is already stripped of
 /// whitespace, comment lines and blank lines at the beginning of the message.
 ///
-pub fn check_stripped_message(message: &str) -> Result<()> {
-    let mut lines = message.lines();
+pub fn check_stripped_message(lines: &mut Iterator<Item = &str>) -> Result<()> {
```
matthiasbeyer: Why not as extension trait for all `Iterator<Item = &str>`?

neithernut: Good point.

neithernut:
As soon as #49 is merged, I can rebase this PR and implement the command.

---

In the (near) future, if we generalize `Lines` and `Trailers`, we can add functions to `LineIteratorExt` returning instances of those. We could also add a function `stripped()` which builds an iterator for stripping a message out of the iterators @matthiasbeyer introduced previously. This way, `LineIteratorExt` may become a primary interface to the `message` module.

lib/src/message/mod.rs
```diff
@@ -7,6 +7,61 @@
...
+    /// message which should not be stored, e.g. commetents and trailing
+    /// whitespace.
+    ///
+    /// Note that the iterator does not (yet) strip blank lines at the beginning
+    /// or end of a message.
+    ///
+    fn stripped(self) -> StripWhiteSpaceRightIter<WithoutCommentsIter<Self::Iter>>;
+}
+
+impl<L> LineIteratorExt for L where L: Iterator<Item = String> {
```
matthiasbeyer: That looks surprisingly ugly. Please consider one of the following:

```rust
impl<L> LineIteratorExt for L
    where L: ...
{
```

or

```rust
impl<L: Iterator<...> LineIteratorExt for L
```

lib/src/message/mod.rs
```diff
@@ -7,6 +7,61 @@
...
+        }
+
+        if !self.next().map(|line| line.is_empty()).unwrap_or(true) {
+            return Err(Error::from_kind(EK::MalformedMessage));
+        }
+
+        Ok(())
+    }
+
+    fn stripped(self) -> StripWhiteSpaceRightIter<WithoutCommentsIter<Self::Iter>> {
```
matthiasbeyer: I don't like the return type signature. But I don't know what the options are here.

src/main.rs
```diff
@@ -94,6 +96,23 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
 
+/// check-message subcommand implementation
+///
+fn check_message(matches: &clap::ArgMatches) -> i32 {
+    let reader: Box<Read> = match matches.value_of("filename") {
+        Some(filename)  => Box::from(try_or_1!(File::open(filename))),
+        None            => Box::from(io::stdin()),
+    };
+    BufReader::new(reader).lines()
+                          .map(|l| l.unwrap())
```
matthiasbeyer: That looks dangerous here.

neithernut: The only way the `unwrap()` panics is on an IO error. It's valid to abort the program in such cases, IMO. We also could throw in an explicit `exit()` instead as an alternative.

lib/src/message/mod.rs
neithernut: Yeah, and I even did not include the `SkipWhile` which I intended originally. I'll guess we have to wait until we can simply return an `Iterator<Item = String>`.

lib/src/message/mod.rs
neithernut: Will do.

PR #56, opened 2017-03-19
Impl of "show" command

matthiasbeyer:
This is ugly, monolithic, copy-pasta-code and does not even compile.

But I open this PR anyways for tracking.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,8 +17,8 @@ mod error;
 mod programs;
 mod util;
 
-use clap::App;
-use git2::{Commit, Oid, Repository};
+use clap::{App, Values};
+use git2::{Commit, Oid, Repository, References};
 use libgitdit::iter::IssueMessagesIter;
 use libgitdit::message::{CommitExt, LineIteratorExt};
 use libgitdit::repository::RepositoryExt;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -95,6 +96,175 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
 
+/// show subcommand implementation
+fn show(repo: &Repository, sub: &clap::ArgMatches) -> i32 {
+    let issue = match sub.value_of("parent") {
+        Some(s) => s,
+        None => {
+            error!("No parent hash");
+            return 1;
+        },
+    };
```
neithernut: If configured correctly, clap should generate an error. An `unwrap` should be sufficient.

neithernut:
You could start feeding leaves of an issue to a `RevWalk` and plotting a simple graph using an iterator pattern. With a few tricks, you can transform a `RevWalk` to a lines iterator which gives you a graph to print.
You could then extend that tree printing iterator and have each item of that iterator expand to a message, including the graph portion at the beginning of the line.

neithernut:
Please rebase and remove the bash version of the subcommand.

matthiasbeyer:
As soon as this is in a shape where one can use it.

PR #61, opened 2017-03-23
Port get-issue-metadata

### LINKED ISSUE #45: Port "get-issue-metadata" to rust
Add a function to `RepositoryExt` for extracting all metadata (or a defined subset) for an issue. Implement the subcommand based on that function.

neithernut:
Still needs some testing. Resolves #45.


neithernut:
Aaaaand it's bust. The regular expression introduces way back isn't quite right.

neithernut:
I fixed the regex. It was foolish to not write tests for the message module (as it is one of the few things that can be tested without a repository).

src/main.rs
```diff
@@ -133,15 +111,26 @@ fn create_message(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
 }
 
 
+/// get-issue-metadata subcommand implementation
+///
+fn get_issue_metadata(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
+    // note: "head" is always present since it is a required parameter
+    let commits = try_or_1!(repo.value_to_commit(matches.value_of("head").unwrap())
+                                .map(|commit| IssueMessagesIter::new(commit, repo)));
+    for trailer in commits.flat_map(|commit| CommitExt::trailers(&commit)) {
```
matthiasbeyer: Why not `commit.trailers()` here? Or `flat_map(CommitExt::trailers)` ?

neithernut: `flat_map(CommitExt::trailers)` doesn't work because we pass a ref. I guess I expanded it to test whether that was the only problem and forgot to write it as `commit.trailers()`.

PR #64, opened 2017-03-24
Add function for preparing pager invocation

### LINKED ISSUE #55: Add utility function for invoking a pager
Some commands will spawn a pager. We should have a function doing this. Like with the editor (#33), the function should honour the git-config. Blocks #41 and #51.

----

It makes sense to rename the `editor` module to `programs` or something similar, then add the pager functionality.

neithernut:
Resolves #55.


src/error.rs
```diff
@@ -29,9 +29,9 @@ error_chain! {
             display("Could not find config: {}", configname)
         }
 
-        EditorError {
-            description("Could not find editor configuration or ENV variable")
-            display("Could not find editor configuration or ENV variable")
+        ProgramError(commandname: String) {
```
matthiasbeyer: `commandname` is misleading.

src/programs.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// A bare command will be returned containing
+///
+fn program(config: Config, command: CommandSpec) -> Result<Command> {
+    config
+        .get_entry(command.config_name())
+        .chain_err(|| EK::ConfigError(command.config_name().to_owned()))?
+        .value()
+        .map(String::from)
+        .or_else(|| command.env_var_names().map(env_var).filter_map(|res| res.ok()).next())
```
matthiasbeyer: `.filter_map(Result::ok)`!

src/programs.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// Get the names of the environment variables which may contain the program
+    /// name
+    ///
+    fn env_var_names(&self) -> slice::Iter<'static, &'static str> {
+        match self {
+            &CommandSpec::Editor => {static X : &'static [&'static str] = &["GIT_EDITOR", "EDITOR"]; X},
+            &CommandSpec::Pager  => {static X : &'static [&'static str] = &["GIT_PAGER",  "PAGER" ]; X},
+        }.into_iter()
```
matthiasbeyer: Thanks for _not_ making this `pub` :smile:

src/programs.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+fn program(config: Config, command: CommandSpec) -> Result<Command> {
+    config
+        .get_entry(command.config_name())
+        .chain_err(|| EK::ConfigError(command.config_name().to_owned()))?
+        .value()
+        .map(String::from)
+        .or_else(|| command.env_var_names().map(env_var).filter_map(|res| res.ok()).next())
+        .map(Command::new)
+        .ok_or_else(|| Error::from(EK::ProgramError(command.name().to_owned())))
```
matthiasbeyer: I beleive this can be written as `EK::ProgrammError(command.name().to_owned()).into()` - but I'm not sure.

src/programs.rs
neithernut: Didn't work for some reason. And I was too lazy to investigate why for the moment. I'll have another look.

neithernut: The compiler reports a type mismatch for the error type when just typing `Result::ok`. I guess that's because of the `Result` type introduced by `error_chain`. We could write `.filter_map(<module-path>::Result::ok)`, but the closure we have now is way shorter and may even be easier to read.

matthiasbeyer: Nope. Feel free to ff-merge https://github.com/neithernut/git-dit/pull/65

PR #67, opened 2017-03-28
Port git-dit-new

neithernut:
Resolves #37.

PR #68, opened 2017-03-28
Additions to the message module

neithernut:
This PR introduces some new interfaces to the `message` module of the library and refactors a few things on the way. In preparation for #67.

### DIFF
--- a/lib/src/error.rs
+++ b/lib/src/error.rs
@@ -40,6 +40,11 @@ error_chain! {
             display("Malformed head refernece: {}", name)
         }
 
+        TrailerFormatError(trailer: String) {
+            description("Found malformed trailer")
+            display("Malformed trailer: {}", trailer)
+        }
+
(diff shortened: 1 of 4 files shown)


neithernut:
I have yet to test whether these changes break anything.

neithernut:
Tested superficially. Should work.

lib/src/error.rs
```diff
@@ -40,6 +40,11 @@ error_chain! {
             display("Malformed head refernece: {}", name)
         }
 
+        TrailerFormatError {
```
matthiasbeyer: Passing in the trailer here would be nice, so we can display it!

neithernut: You mean the string which we tried to parse into a trailer.

matthiasbeyer: Yep, exactly.

PR #70, opened 2017-03-30
Port git-dit-reply

### LINKED ISSUE #38: Port "git-dit-reply" to rust
What the title says. Porcelain commands should go into the binary.

neithernut:
Will, at some point, resolve #38.


src/main.rs
```diff
@@ -182,6 +182,88 @@ fn new_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+            format!("Re: {}", s)
+        }
+    });
+    let tree = try_or_1!(parent.tree());
+
+    // figure out to what issue we reply
+    // NOTE: The following is this ugly because `Clone` is not implemented for
+    //       `git2::Commit`, `Debug` is not implemented for `git2::Object` and
+    //       `find_tree_init()` consumes the commit because of internals.
+    let issue = try_or_1!(repo.find_tree_init(parent.as_object().clone().into_commit().ok().unwrap())).id();
```
neithernut: Maybe it makes sense to have `find_tree_init()` take a reference (and move the ugly workaround over there). It would clearly make sense regarding semantics. The sometimes unnecessary clone would be a drawback, but this operation _should_ be relatively cheap.

I'm open to suggestions here.

neithernut:
Still needs to be tested.

neithernut:
Tested superficially. Appears to work.

src/main.rs
```diff
@@ -182,6 +182,85 @@ fn new_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+
+            try_or_1!(file.consume_lines(try_or_1!(repo.prepare_trailers(matches))));
+            try_or_1!(file.flush());
+        }
+
+        try_or_1!(repo.get_commit_msg(path))
+    }.into_iter().collect_string();
+
+    // construct a vector holding all parents
+    let parent_refs : Vec<&Commit> = Some(&parent).into_iter().chain(references.iter().map(|command| command)).collect();
```
matthiasbeyer: `.map(|c| c)`?

matthiasbeyer: I just compiled this branch without the unecessary `map`. Remove it and I approve. :+1:

PR #75, opened 2017-04-01
Fix command line

neithernut:
There are multiple issues with the `src/cli.yaml`. Some settings are wrong, so are some strings in the source. We should have checked them earlier. This PR is intended for cleaning up the mess that we shouldn't have produced in the first place.

Most notably, the `multiple` setting for options appears to make clap account any argument following the option, until the next one starting with a dash `-`, to the option, even if a required parameter is to be expected. As a work-around, one can insert a `--`, but the behaviour is not what one wants and confusing to users, imo.

### DIFF
--- a/src/cli.yaml
+++ b/src/cli.yaml
@@ -271,6 +271,7 @@ subcommands:
                 help: Message to which to reply to
                 index: 1
                 multiple: false
+                required: true
 
     - show:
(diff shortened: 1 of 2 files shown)


neithernut:
More commits will follow. I'd appreciate hints, if someone spots/encounters something which ought to be fixed in this PR.

PR #80, opened 2017-04-06
Port git-dit-fetch

### LINKED ISSUE #52: Port "git-dit-fetch" to rust
What the title says.

neithernut:
Resolves #52.


neithernut:
Needs testing.

neithernut:
Doesn't work yet, at least not if authentication is required. I guess we have to provide callbacks.

src/main.rs
```diff
@@ -146,6 +147,41 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+/// fetch subcommand implementation
+///
+fn fetch_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
+    let mut remote = try_or_1!(repo.find_remote(matches.value_of("remote").unwrap()));
+
+    // accumulate the refspecs to fetch
+    let refspecs : Vec<String> = if let Some(issues) = matches.values_of("issue") {
+        // fetch a specific list of issues
+        let iter = issues.map(Oid::from_str)
+                         .abort_on_err();
```
matthiasbeyer: You can put this on one line, I guess.

src/main.rs
```diff
@@ -146,6 +147,41 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+        if matches.is_present("known") {
+            iter.chain(try_or_1!(repo.get_all_issue_hashes()).abort_on_err())
```
matthiasbeyer: Wait - a `try!()` and on the return value we call `abort_on_err()`, which is a function on `Result<_>` - so we basically have a `Result<Result<_>, _>` unwrapping here? That sounds not like something we want!

src/main.rs
```diff
@@ -146,6 +147,41 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+                .map(|issue| remote.issue_refspec(issue).unwrap())
```
matthiasbeyer: `unwrap()` safe here? If yes, please add a comment!

src/main.rs
```diff
@@ -146,6 +147,41 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+                .collect()
+        } else {
+            iter.map(|issue| remote.issue_refspec(issue).unwrap())
```
matthiasbeyer: `unwrap()` safe here? If yes, please add a comment!

src/main.rs
```diff
@@ -146,6 +147,41 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+                .collect()
+        }
+    } else {
+        vec![remote.all_issues_refspec().unwrap()]
```
matthiasbeyer: `unwrap()` safe here? If yes, please add a comment!

src/main.rs
```diff
@@ -146,6 +147,41 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
...
+    };
+
+    // set the options for the fetch
+    let mut fetch_options = FetchOptions::new();
+    fetch_options.prune(match matches.is_present("prune") {
```
matthiasbeyer: You really should change this `match` to `if`!

src/main.rs
neithernut: Nope, it's a `Result<Iterator<Item = Result<_, _>, _>`.

neithernut:
@matthiasbeyer should we use [rpassword](https://github.com/conradkleinespel/rustastic-password) to read credentials for fetching/pushing?

matthiasbeyer:
I don't see why not!

neithernut:
It's surprisingly hard to get ssh authentication working, mostly due to missing documentation in libgit2, but also because there appear to be a few interoperability issues still unresolved. (e.g. the ssh-config not being consulted).

matthiasbeyer:
We could contribute to git2-rs and improve documentation there! :smile:

neithernut:
For the sake of progress, maybe we should only support anonymous access and access using an ssh-agent for now. In the future, we will introduce subscriptions anyway (In fact, we are using manual subscriptions already). Those would cause `git fetch` to fetch issues already and hence do authentication.nAlso, we may improve authentication capabilities in the future.

Cmments? @matthiasbeyer?

matthiasbeyer:
I don't see why we are even talking about authentication here? It is absolutely not our domain AFAICS.

neithernut:
The only thing that's missing right now is notifying the user about new or removed messages (through the `update_tips` callback).

src/main.rs
```diff
@@ -151,6 +153,38 @@ fn get_issue_tree_init_hashes(repo: &Repository, _: &clap::ArgMatches) -> i32 {
 
 // Porcelain subcommand implementations
 
+/// fetch subcommand implementation
+///
+fn fetch_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
+    let mut remote = try_or_1!(repo.find_remote(matches.value_of("remote").unwrap()));
```
matthiasbeyer: A comment that this is safe via clap (the unwrap call) would be nice.

PR #81, opened 2017-04-07
Enable logging

neithernut:
Add a basic logger and wire it up, so we actually see the errors we are reporting.

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -1,0 +1,47 @@
+//   git-dit - the distributed issue tracker for git
+//   Copyright (C) 2016 Matthias Beyer <mail@beyermatthias.de>
+//   Copyright (C) 2016 Julian Ganz <neither@nut.email>
+//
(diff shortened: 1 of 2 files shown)


neithernut:
Currently, if an error is logged, the "wrapped errors" obfuscate inner layers. Maybe we have to adjust our error printing (e.g. the `try_or_1` macro).

neithernut:
The fact that `ChainedError`s should not be printed directly but via the struct returned by `display()` makes things a bit more complicated. Hence, I made this WIP.

src/logger.rs
```diff
@@ -0,0 +1,47 @@
...
+//
+
+use log::{self, LogRecord, LogLevel, LogMetadata};
+use io::{stderr, Write};
+use std::result::Result as RResult;
+
+
+/// Basic logger
+///
+/// This logger will log to stdout
```
matthiasbeyer: This is wrong. It logs to stderr.

PR #86, opened 2017-05-15
Port "show" subcommand to rust

### LINKED ISSUE #51: Port "git-dit-show" to rust
What the title says.

neithernut:
Closes #51
Closes #56

Note: the following has yet to be done in this PR:
 * [x] Differentiate between different kinds of `Mark`s. The differentiation is required because we loose visual tracking information if we replace all marks when expanding graph elem lines.
 * [x] Add module level documentation. I doubt that most people would understand what the code does.
 * [x] Introduce proper commit message formatting.
 * [x] Support `--initial` option (which shows only the initial message).
 * [ ] Introduce a type bundling a `(TreeGraphElemLine, Display)` and impl `Display` for it. This might make things slightly more readable and saves us one copying of a string per line. (optional)


src/msgtree.rs
```diff
@@ -8,17 +8,36 @@
...
+pub enum TreeGraphElem {
     Empty,
     Following,
     Mark,
 }
 
 
+/// Representation of one line of tree graph elements
+///
+pub struct TreeGraphElemLine { elems: Vec<TreeGraphElem> }
```
matthiasbeyer: Why not `pub struct TreeGraphElemLine(Vec<TreeGraphElem>);` ?

neithernut: It simply didn't occur to me.

src/main.rs
```diff
@@ -384,11 +384,26 @@ fn show_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
     let issue = try_or_1!(Oid::from_str(matches.value_of("issue").unwrap()));
-    let graph = try_or_1!(repo.get_issue_revwalk(issue)).abort_on_err()
+    let commits = try_or_1!(repo.get_issue_revwalk(issue)).abort_on_err()
         .map(|oid| repo.find_commit(oid)).abort_on_err()
         .into_tree_graph();
 
-    // TODO: reverse the iterator if necessary (chronological vs. "timeline")
+    // Transform the simple graph element line into an iterator we may use for
+    // zipping with the message, reversing the order if necessary
+    let graph : Box<Iterator<Item=(CommitTreeGraphLines, Commit)>> = if matches.is_present("tree") {
```
matthiasbeyer: Please use whitespace around `=`.

src/main.rs
```diff
@@ -384,11 +384,26 @@ fn show_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+        // We want the commits in chronological order
+        let mut buf : Vec<(TreeGraphElemLine, Commit)> = commits.collect();
+        buf.reverse();
```
matthiasbeyer: We should introduce `itertools` as dependency and use the `rev` iterator from there. This way you don't have to collect here and we can keep the iterator... which might bring us a big benefit later on!

neithernut: I'll look into it.
I already considered something similar, but it would have meant using a `Box<Iterator>`. And if I have the choice between a boxed iterator over possibly pre-collected values and a straight `Vec`, I prefer the `Vec`.

neithernut: Meh.
Just looked up `itertools`. As I suspected, using it would mean boxed iterators. So there's no real benefit. The number of commits should be moderate anyway. Collecting doesn't really hurt.

neithernut:
At this point, `git dit show` displays a graph with commit ids. Only the `-g` option is supported, which reverses the order of commits. Feel free to play around with this and check whether the topology displayed mirrors what `git log --graph` says. It _should_ work, though.

I consider an intermediate squash. The history is a mess right now and given that the thing works as intended, I'll only build upon the current state from now on. Which means (hopefully) no more fixups to commits which are currently pushed. Hence, I'd appreciate a review, @matthiasbeyer.

lib/src/repository.rs
```diff
@@ -109,6 +113,17 @@ impl RepositoryExt for Repository {
...
     }
 
+    fn get_issue_revwalk(&self, issue: Oid) -> Result<Revwalk> {
+        let glob = format!("**/dit/{}/leaves/*", issue);
+        self.revwalk()
+            .and_then(|mut revwalk| {
+                revwalk.push_glob(glob.as_ref())?;
+                revwalk.simplify_first_parent();
+                revwalk.set_sorting(git2::SORT_TOPOLOGICAL);
+                Ok(revwalk) })
```
matthiasbeyer: The `})` should be on the next line.

src/main.rs
```diff
@@ -373,6 +375,55 @@ fn reply_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+fn show_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
+    // NOTE: we use boxed iterators since we might do a lot of dynamic wrapping
+    //       depending on command line parameters
+
+    // first, get us an iterator over all the commits
+    // NOTE: "issue" is a required parameter
+    let issue = try_or_1!(Oid::from_str(matches.value_of("issue").unwrap()));
+    let mut commits : Vec<(TreeGraphElemLine, Commit)> =
+        try_or_1!(repo.get_issue_revwalk(issue)).abort_on_err()
+            .map(|oid| repo.find_commit(oid)).abort_on_err()
```
matthiasbeyer: The `abort_on_err()` calls above should be on individual lines.

src/main.rs
```diff
@@ -373,6 +375,55 @@ fn reply_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+            .collect();
+
+    // Decide on the order in which the messages will be printed.
+    if matches.is_present("tree") {
+        // We want the commits in chronological order
+        commits.reverse();
+        for commit in commits.iter_mut() {
+            commit.0.reverse_marks();
+        }
+    };
```
matthiasbeyer: I'm not entirely sure whether this is a nice idea. We're introducing a _lot_ of overhead with this reverse thing here (because we first `collect` and then build a new iterator from it).

Using the reverse functionality from `itertools`, this would be _way_ more performant (no `collect`) and LLVM would be able to optimize this _way_ more agressively.

But, on the other hand, we have the problem that introducing a new iterator here for doing this is also a lot of code.

### Suggestion

I would like to fork your PR and optimize this, if you ACK me on this.

src/main.rs
```diff
@@ -373,6 +375,55 @@ fn reply_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+            elems.append(TreeGraphElem::Empty);
+            (elems.commit_iterator(), commit.1)
+        })
+        // expand the message to a series of lines
+        .flat_map(|commit| commit.0.zip(vec![commit.1.id().to_string(), String::new()]))
+        // combine each line of graph elements and message
+        .map(|line| format!("{} {}", line.0, line.1));
+
+    // spawn a pager and write the graph
+    let mut pager = try_or_1!(programs::pager(try_or_1!(repo.config())));
```
matthiasbeyer: Maybe we should open the pager object before the whole iterator chain, for early-fail reasons.

src/main.rs
```diff
@@ -373,6 +375,55 @@ fn reply_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
...
+    try_or_1!(pager.stdin.as_mut().unwrap().consume_lines(graph));
+
+    // don't trash the shell by exitting with a child still printing to it
+    try_or_1!(pager.wait()).code().unwrap_or(1)
```
matthiasbeyer: This is a bit hard to read, IMO. Please put `.code().unwrap_or(1)` each on new lines.

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.0.push(e);
+    }
+
+    /// Reverse "start" and "end" marks
+    ///
+    pub fn reverse_marks(&mut self) {
+        for elem in self.0.iter_mut() {
+            match elem {
+                &mut TreeGraphElem::Mark(MarkType::Start) => *elem = TreeGraphElem::Mark(MarkType::End),
+                &mut TreeGraphElem::Mark(MarkType::End) => *elem = TreeGraphElem::Mark(MarkType::Start),
```
matthiasbeyer: Why not implementing a `fn reverse(self) -> MarkType` on `MarkType` for less complex-to-read code?

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn commit_iterator(self) -> CommitTreeGraphLines {
+        CommitTreeGraphLines(self.0)
+    }
+}
+
+impl FromIterator<TreeGraphElem> for TreeGraphElemLine {
+    fn from_iter<I>(iter: I) -> Self
+        where I: IntoIterator<Item=TreeGraphElem>
+    { TreeGraphElemLine(iter.into_iter().collect()) }
```
matthiasbeyer: `collect`ing here is a performance loss AF.

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+impl fmt::Display for TreeGraphElemLine {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        let mut has_mark = false;
+        for elem in self.0.iter() {
+            f.write_char(match (has_mark, elem) {
+                (_,     &TreeGraphElem::Mark(_)  ) => {has_mark = true; '+'},
+                (true,  _                        ) => '-',
+                (false, &TreeGraphElem::Empty    ) => ' ',
+                (false, &TreeGraphElem::Following) => '|',
+            })?;
```
matthiasbeyer: Suggestion:

* Add `is-match` as dependency
* Implementing `to_str()` or something similar on `TreeGraphElem`
* Implementing this whole match like this:

```
{
    if is_match!(elem, &TreeGraphElem::Mark(_)) { has_mark = true; }
    if has_mark { '-' } else { elem.to_s() }
}
```

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+        for elem in self.0.iter_mut() {
+            match elem {
+                &mut TreeGraphElem::Mark(MarkType::End) => *elem = TreeGraphElem::Empty,
+                &mut TreeGraphElem::Mark(_) => *elem = TreeGraphElem::Following,
+                _ => {},
+            }
+        }
+
+        Some(retval)
+    }
```
matthiasbeyer: Wait. You're cloning `self.0` here, then iterate over the _old_ value and mutating it, then returning the cloned _unmodified_ copy?

If this is actually what you want, I'd expect a **big** comment why this is necessary or can't be done in another way, as it seems to be a huge performance bottleneck here.

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+            let mut elems : TreeGraphElemLine = self.parents.iter_mut().map(
+                |mut parent| match *parent {
+                    Some(id) => if commit.id() == id {
+                            // the current commit is a parent we were awaiting
+                            *parent = parent_update.take();
+                            TreeGraphElem::Mark(if parent.is_some() {MarkType::Mid} else {MarkType::End})
+                        } else { TreeGraphElem::Following },
+                    None => TreeGraphElem::Empty,
+                }
+            ).collect();
```
matthiasbeyer: Please add a scope here:

```
.map(|mut parent| { // ...
})
```

this looks really ugly this way.

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+            // nice horizontal merges in our graph.
+            let mut parent_update = commit.parent(0).as_ref().map(Commit::id).ok();
+
+            // generate graph elements for the parents currently tracked
+            let mut elems : TreeGraphElemLine = self.parents.iter_mut().map(
+                |mut parent| match *parent {
+                    Some(id) => if commit.id() == id {
+                            // the current commit is a parent we were awaiting
+                            *parent = parent_update.take();
+                            TreeGraphElem::Mark(if parent.is_some() {MarkType::Mid} else {MarkType::End})
```
matthiasbeyer: I would add a variable for the mark type here, this is unreadable.

src/msgtree.rs
```diff
@@ -0,0 +1,178 @@
...
+            if parent_update.is_some() {
+                // a history for the current commit doesn't yet exist, so we
+                // should create one
+                elems.append(TreeGraphElem::Mark(MarkType::Start));
+                self.parents.push(parent_update);
+            }
+
+            // keep the graph slim by removing parent trackig information that
+            // is no longer required from the back of the list
+            while self.parents.last().map(|p| p.is_none()).unwrap_or(false) {
```
matthiasbeyer: Simple way to improve readability here: `.map(Option::is_none).unwrap_or(false)`!

src/main.rs
neithernut: If we spawn a pager early and then produce errors (which are printed to `stderr`), we may trash teh terminal.

src/main.rs
neithernut: Note: it's excactly the same line as in the `list_impl()` function.

src/msgtree.rs
neithernut: I doubt it would be much easier to read, but it isn't a bad idea either.

src/msgtree.rs
neithernut: Suggestions?

src/msgtree.rs
neithernut: Ah, I misread your suggestion. Yes, that would probably help.

src/msgtree.rs
neithernut: It is necessary for the commit marks to be displayed correctly. we have to replace the `Mark`s with `Following` after we returned the first line of graph elements, so the mark will only be displayed at the first line for each commit.
I _could_ use an internal state of some sort.

src/msgtree.rs
matthiasbeyer: Storing the `Iterator` object inside `TreeGraphElemLine` would be the best way, though that might complicate some things. But always better than `collect`ing!

neithernut: We'll refactor later.

src/msgtree.rs
neithernut: Not worth doing it now. Maybe we'll refactor later[tm].

neithernut:
So much for "no more fixup commits". But having the documentation introduced in the same commit as the data structure makes sense.

neithernut:
I'll implement neither the `--verify-gpg` nor the `--format` option in this PR: 
I'm not sure whether to add support for the `--decorate` flag.
Opinions? @matthiasbeyer?

neithernut:
Also: I'd opt for implementing a convenience type for `(TreeGraphElemLine, Display)` later.

PR #89, opened 2017-05-29
Add more documentation to library

### LINKED ISSUE #76: Make sure everything in the lib is documented before release
We should really ensure everything is documented (`#![deny(missing_docs)]`), even better would be:

```
#![deny(
    non_camel_case_types,
    non_snake_case,
    path_statements,
    trivial_numeric_casts,
    unstable_features,
    unused_allocation,
    unused_import_braces,
    unused_imports,
    unused_mut,
    unused_qualifications,
    while_true,
)]
```

before releasing.

neithernut:
In this PR, I add missing documentation to stuff in the library.
Resolves #76.

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -7,6 +7,32 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 //
 
+//! The git-dit library
+//!
+//! This library provides low-level functionality for accessing, creating and
+//! manipulating "git-dit" issues and messages. It is implemented on top of the
+//! `git2` crate. This librarie's documentation primarily provides information
+//! about its API and abstract processing of issues and messages.
+//!
+//!
+//! # Issues
+//!
+//! Issues are stored in git repositories. The issues availible in a repository
+//! may be accessed through the `RepositoryExt` extension trait implementation
+//! for `git2::Repository`.
+//!
+//! An issue is primarily a tree of messages, consisting of at least an initial
+//! message. An issue also has a "head reference". The head reference lets the
+//! maintainer indicate an "upstream status" of the issue, e.g. by pointing to a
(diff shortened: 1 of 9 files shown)


neithernut:
The library should be mostly done. The one thing that is missing documentation should be the `lib.rs`. I'm not quite sure what approach to take:
 * We could move the existing documentation in `doc` to `lib`. It would enrich the library documentation and we may still refer to it from the `README.md`. However, I don't want to mix the specification and the (reference-)implementation. We might even create a new repository for the specification itself at some point. Also, having the documentation visible at the root has its perks, too.
 * We could add a lengthy introduction to `lib.rs`. We would end up duplicating much of the content form the specification. Alternatively, we might as well just refer to the spec and try to keep the crate documentation small.

----

I'm also not sure whether we should include a `#![deny(missing_docs)]` in the code at this stage. We'll definitely refactor rather soonish and having to add dummy documentation for every little experiment is rather cumbersome. Passing `-D missing_docs` to `rustc` in the CI or the `-A` flag locally would be an option, but sadly it appears that you cannot specify them when using `cargo`.

neithernut:
Also: I often used inline code-snippets when referring to structs and traits in documentation. Maybe there is a way to refer to them more explicitly but I didn't yet find out how to.

lib/src/lib.rs
```diff
@@ -7,6 +7,16 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 //
 
+//! The git-dit library
+//!
+//! This library provides low-level functionality for accessing, creating and
+//! manipulating "git-dit" issues and messages. This librarie's documentation
+//! primarily provides information about its API and abstract processing of
+//! issues and messages. In the future, we will also provide a detailed
```
matthiasbeyer: I would not promise what we might not can keep.

PR #91, opened 2017-05-30
Enable old rustc versions on travis

matthiasbeyer:
The reason for this is that some distributions are rather old for shipping updates here and (as far as I know by now) we can compile this with old versions.

If something breaks in an old version, we can fix is before the PR lands, as far as it is not too complicated to fix. Of course, the current version(s) of the rust compiler are prefered, but in most situations fixing the codebase for older versions won't hurt that much and the benefits are clearly more important than the draw-backs.

---

But before everything we should see whether this succeeds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ rust:
   - stable
   - beta
   - nightly
+  - 1.13.0
+  - 1.14.0
+  - 1.15.0
+  - 1.16.0
 matrix:
   allow_failures:
     - rust: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 language: rust
 rust:
+  - 1.15.0
+  - 1.16.0
   - stable
   - beta
   - nightly
-  - 1.13.0
-  - 1.14.0
-  - 1.15.0
-  - 1.16.0
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


neithernut:
Aaaaaaand it's bust.

neithernut:
The reason for the current failure appears to be some dependency. Maybe the error goes away if we downgrade.

matthiasbeyer:
[Rust 1.14 came out on 2016-12-22](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1140-2016-12-22), which we should support imo. 

Debian ships rustc 1.7 as far as I can see. I _guess_ we _could_ support this version, but I'm fairly sure that it will be pita. I will see whether I can fix things up to support debian builds. But I won't guarantee anything. I will do more research on this topic in the next two weeks, if that's okay with you!

neithernut:
I guess we should support Rust 1.14, but I wouldn't put much effort into supporting Rust 1.7 out of the box, from the master. I would rather declare that a case for explicit backporting.

matthiasbeyer:
Limited support to rustc >= 1.15. Rationale in the message of the latest commit.

matthiasbeyer:
I'd like to preserve this piece of commit history. Feel free to push the button.

Issue #93, opened 2017-06-01
Remove the Makefile


neithernut:
The `Makefile` is no longer used.

PR #94, opened 2017-06-01
Make Cargo build the manpage

### LINKED ISSUE #92: Build and install man page
There are two possible solutions:
 * make cargo build the existing man page (`git-dit.1.md`)
 * build a man page from the `clap` CLI spec

Blocks #93.

neithernut:
This PR makes Cargo build the man page for us. The installation is, however, not performed. Since the `Makefile` was only introduced in order to build the man page, it was removed.
Resolves #92 and #93.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,4 @@ addons:
     packages:
       - cmake
       - cmake-data
+      - pandoc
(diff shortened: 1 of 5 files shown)


neithernut:
Meh. Need to install pandoc on the CI server.

neithernut:
Should we make the man page an optional (but default) feature?

neithernut:
Meh, fails on `rust-1.15` because the `Command` interface changed.

PR #95, opened 2017-06-01
Remove WrappedGitError and replace with explicit error kinds

matthiasbeyer:

PR #96, opened 2017-06-01
Execute buildscript only if desired

matthiasbeyer:
The problem is that I don't want to install pandoc for development... and we cannot expect the casual hacker to do this, either.

So we make this dependend on a env variable. At least this is what I propose.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,5 @@ addons:
       - cmake
       - cmake-data
       - pandoc
+env:
+  - BUILD_GIT_DIT_MAN=1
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Sidenote: I need this to finish my work on #95 (fixing the WrappedGitError thing in the CLI app as well).

neithernut:
I get the impression that Cargo is rather limited compared to other build environments (CMake, autotools, ...)

matthiasbeyer:
You only get that impression because you "need" features nobody else needs.

PR #97, opened 2017-06-01
Add a dedicated type for issues

neithernut:
Blocks #99.

### DIFF
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -39,6 +39,16 @@ impl<'r> Issue<'r> {
     pub fn id(&self) -> Oid {
         self.id
     }
+
+    /// Get reference part unique for this issue
+    ///
+    /// The references associated with an issue reside in paths specific to the
+    /// issue. This function returns the part unique for the issue, e.g. the
+    /// part after the  `dit/`.
+    ///
+    fn unique_ref_part(&self) -> String {
+        self.id.to_string()
+    }
 }
 
 impl<'r> fmt::Display for Issue<'r> {
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -67,6 +67,18 @@ impl<'r> Issue<'r> {
             .chain_err(|| EK::CannotFindIssueHead(self.id))
     }
 
+    /// Get the leaf references for the issue
+    ///
+    /// Returns the leaf references for the issue from both the local repository
+    /// and remotes.
+    ///
+    pub fn issue_leaves(&self) -> Result<References<'r>> {
+        let glob = format!("**/dit/{}/leaves/*", self.unique_ref_part());
+        self.repo
+            .references_glob(&glob)
+            .chain_err(|| EK::CannotGetReferences(glob))
+    }
+
     /// Get reference part unique for this issue
     ///
     /// The references associated with an issue reside in paths specific to the
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -12,7 +12,7 @@
 //! This module provides the `Issue` type and related functionality.
 //!
 
-use git2::{self, Oid, References};
+use git2::{self, Oid, Reference, References};
 use std::fmt;
 use std::result::Result as RResult;
 
@@ -52,6 +52,18 @@ impl<'r> Issue<'r> {
         let glob = format!("**/dit/{}/head", self.unique_ref_part());
         self.repo
             .references_glob(&glob)
+            .chain_err(|| EK::CannotFindIssueHead(self.id))
+    }
+
+    /// Get the local issue head for the issue
+    ///
+    /// Returns the head reference of the issue from the local repository, if
+    /// present.
+    ///
+    pub fn find_local_head(&self) -> Result<Reference<'r>> {
+        let refname = format!("refs/dit/{}/head", self.unique_ref_part());
+        self.repo
+            .find_reference(&refname)
             .chain_err(|| EK::CannotFindIssueHead(self.id))
     }
 
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -79,6 +79,18 @@ impl<'r> Issue<'r> {
             .chain_err(|| EK::CannotGetReferences(glob))
     }
 
+    /// Get all local references for the issue
+    ///
+    /// Return all references associated with the issue from the local
+    /// repository.
+    ///
+    pub fn local_refs(&self) -> Result<References<'r>> {
+        let glob = format!("refs/dit/{}/**", self.unique_ref_part());
+        self.repo
+            .references_glob(&glob)
+            .chain_err(|| EK::CannotGetReferences(glob))
+    }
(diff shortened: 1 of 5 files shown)


neithernut:
Mh... originally I planned on creating the new `Issue` type in the root module of the library and moving the repository over, too. Now I'm not so sure any more whether having the `Issue` in the root module is such a good idea. Issues are virtually always queried from the repository anyway so the complete module path may not show up as often as I anticipated.
Also, an `issue` could also contain some issue-specific helper types (in the same file).
Any thoughts, @matthiasbeyer?

neithernut:
Moving the `Issue` type into a new module had the effect that `Issue::new` has to be public for `repository::find_issue()` to compile. I would have liked to not expose that function.

lib/src/issue.rs
```diff
@@ -0,0 +1,128 @@
...
+        self.id
+    }
+
+    /// Get possible heads of the issue
+    ///
+    /// Returns the head references from both the local repository and remotes
+    /// for this issue.
+    ///
+    pub fn heads(&self) -> Result<References<'r>> {
+        let glob = format!("**/dit/{}/head", self);
```
matthiasbeyer: Don't use `self` here for formatting! If the impl of `Display` changes, this will result in strange errors which are hard to debug. Explicit > Implicit!

lib/src/issue.rs
```diff
@@ -0,0 +1,128 @@
...
+            .chain_err(|| EK::WrappedGitError)
+    }
+
+    /// Get the local issue head for the issue
+    ///
+    /// Returns the head reference of the issue from the local repository, if
+    /// present.
+    ///
+    pub fn find_local_head(&self) -> Result<Reference<'r>> {
+        let refname = format!("refs/dit/{}/head", self);
```
matthiasbeyer: Same here.

lib/src/issue.rs
```diff
@@ -0,0 +1,128 @@
...
+            .chain_err(|| EK::WrappedGitError)
+    }
+
+    /// Get the leaf references for the issue
+    ///
+    /// Returns the leaf references for the issue from both the local repository
+    /// and remotes.
+    ///
+    pub fn issue_leaves(&self) -> Result<References<'r>> {
+        let glob = format!("**/dit/{}/leaves/*", self);
```
matthiasbeyer: And here...

lib/src/issue.rs
```diff
@@ -0,0 +1,128 @@
...
+            .chain_err(|| EK::WrappedGitError)
+    }
+
+    /// Get all local references for the issue
+    ///
+    /// Return all references associated with the issue from the local
+    /// repository.
+    ///
+    pub fn local_refs(&self) -> Result<References<'r>> {
+        let glob = format!("refs/dit/{}/**", self);
```
matthiasbeyer: And here.

lib/src/issue.rs
```diff
@@ -0,0 +1,128 @@
...
+            .references_glob(&glob)
+            .chain_err(|| EK::WrappedGitError)
+    }
+
+    /// Get a revwalk for traversing all messages of the issue
+    ///
+    /// The sorting of the revwalk will be set to "topological".
+    ///
+    pub fn message_revwalk(&self) -> Result<git2::Revwalk<'r>> {
+        let glob = format!("**/dit/{}/**", self);
```
matthiasbeyer: And here.

src/main.rs
```diff
@@ -412,7 +416,7 @@ fn show_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
         if matches.is_present("initial") {
             vec![(TreeGraphElemLine::empty(), try_or_1!(repo.find_commit(issue)))]
         } else {
-            try_or_1!(repo.get_issue_revwalk(issue))
+            try_or_1!(repo.find_issue(issue).and_then(|issue| issue.message_revwalk()))
```
matthiasbeyer: You can put the `and_then()` call on a new line here.

Also (I'm not sure with this one) you could use `and_then(Issue::message_revwalk)` which would result in a bit cleaner code.

src/main.rs
```diff
@@ -463,7 +467,7 @@ fn tag_impl(repo: &Repository, matches: &clap::ArgMatches) -> i32 {
     let issue = try_or_1!(Oid::from_str(matches.value_of("issue-hash").unwrap()));
 
     // get the head for the issue to tag
-    let mut issue_head = try_or_1!(repo.get_local_issue_head(issue));
+    let mut issue_head = try_or_1!(repo.find_issue(issue).and_then(|issue| issue.find_local_head()));
```
matthiasbeyer: Same points as above (the first one only if you care, because everything fits in one line, still).

lib/src/issue.rs
neithernut: It is meant to do just that. The rationale is that the implementation of `Display` always results in a unique string which may be used as part of reference names. This way, if we introduce "scoping" or "namespacing" for issues, the scope or namespace finds its way into references automatically.

We also could have a function which explicitly returns a string referring to the issue, which is specified to be used in ref-names and have the `Display` implementation rely on that function instead. If you insist, that is.

src/main.rs
neithernut: Not quite. You have to get a reference first.

lib/src/issue.rs
matthiasbeyer: The problem I'm talking about is that `Display` is explicitely for returning a string which gets shown to the user:

> Display is similar to Debug, but Display is for user-facing output

(from [here](https://doc.rust-lang.org/std/fmt/trait.Display.html))

Therefor, it's implementation may change to return something like

```
"Issue d630a948079bee0adc5f"
```

which would result in huge bugs.

It is not intended to return an identifier or something like that.

neithernut:
I squashed and rebased on the new master, with new error kinds.

lib/src/issue.rs
```diff
@@ -0,0 +1,138 @@
...
+            .and_then(|mut revwalk| {
+                // The iterator will iterate over all the messages in the tree
+                // spanned but it will halt at the initial message.
+                revwalk.push_glob(glob.as_ref())?;
+                self.repo
+                    .find_commit(self.id)
+                    .and_then(|commit| commit.parent_id(0))
+                    .ok() // the initial message having no parent is not unusual
+                    .map(|parent| revwalk.hide(parent))
+                    .unwrap_or(Ok(()))?;
```
matthiasbeyer: I find that line a bit hard to follow (as in "why a default value here... we do not use it"). That would be clearer if we'd use a `let _ = self.repo` at the beginnging of this block of code... just a readability thing!

src/main.rs
matthiasbeyer: Either way you can put the `and_then` on a new line here.

src/main.rs
matthiasbeyer: ...and here.

PR #98, opened 2017-06-02
Consolidate line processing iterators

neithernut:
This PR consolidates line processing iterators by:
 * moving line processing iterators to the message module, where they belong (because they are used for processing messages)
 * getting rid of the unnecessary convenience traits for creating those iterators (the `LineIteratorExt` is the convenience trait)


lib/src/message/line_processor.rs
```diff
@@ -0,0 +1,143 @@
...
+          S: AsRef<str>
+{
+    type Item = String;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next().map(|l| {
+            let line = l.as_ref();
+            match line.is_empty() {
+                true  => String::from(">"),
+                false => String::from("> ").add(line),
```
matthiasbeyer: I'm not sure whether `format!("> {}", line)` would result in faster code here. Care to do some research?

lib/src/message/line_processor.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn from(lines: I) -> Self {
+        StripWhiteSpaceLeftIter(lines)
+    }
+}
+
+impl<'a, I, S> Iterator for StripWhiteSpaceLeftIter<I, S>
+    where I: Iterator<Item = S> + Sized,
+          S: AsRef<str>
+{
+    type Item = String;
```
matthiasbeyer: Do we need to use `String` here as an Item? Couldn't we use `AsRef<str>` as well?

lib/src/message/line_processor.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn from(lines: I) -> Self {
+        StripWhiteSpaceRightIter(lines)
+    }
+}
+
+impl<'a, I, S> Iterator for StripWhiteSpaceRightIter<I, S>
+    where I: Iterator<Item = S> + Sized,
+          S: AsRef<str>
+{
+    type Item = String;
```
matthiasbeyer: Same comment as above.

lib/src/message/line_processor.rs
```diff
@@ -0,0 +1,143 @@
...
+impl<I, S> Iterator for WithoutCommentsIter<I, S>
+    where I: Iterator<Item = S> + Sized,
+          S: AsRef<str>
+{
+    type Item = S;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(next) = self.0.next() {
+            // we do not trim whitespace here, because of code blocks in the message which might
+            // have a "#" at the beginning
```
matthiasbeyer: Thanks for this comment! :+1:

lib/src/message/line_processor.rs
neithernut: Good point. It may not result in less copying of strings in our code, though (because the lift). At least not yet.

lib/src/message/line_processor.rs
neithernut: I'll check that.

lib/src/message/line_processor.rs
matthiasbeyer: But it results in cleaner code, as a user of the iterator is more flexible with types.

lib/src/message/line_processor.rs
neithernut: As suspected: `format!` is twice as fast.

lib/src/message/line_processor.rs
neithernut: Well, we can't use `asRef<str>` as an Item type, since it's a trait.

matthiasbeyer: Can we use `&str`? Would still be cleaner interface-wise, IMO.

neithernut: For now, I'll not chage this, unless either of us comes up with an idea.

PR #100, opened 2017-06-07
Get rid of the FirstParentIter

neithernut:
Use a `Revwalk` instead, and simplify things by providing convenience functions through the `RepositoryExt`.


lib/src/iter.rs
```diff
@@ -61,36 +60,38 @@ impl<'r> Iterator for HeadRefsToIssuesIter<'r>
...
 }
 
 impl<'r> IssueMessagesIter<'r> {
-    pub fn new<'a>(commit: Commit<'a>, repo: &'a Repository) -> IssueMessagesIter<'a> {
-        IssueMessagesIter {
-            inner: FirstParentIter::new(commit),
-            repo: repo,
+    pub fn new<'a>(repo: &'a Repository, commit: Commit<'a>) -> Result<IssueMessagesIter<'a>> {
+        let revwalk = repo.first_parent_revwalk(commit.id())?;
+        Ok(IssueMessagesIter { inner: revwalk, repo: repo })
```
matthiasbeyer: Why not 

```rust
repo
    .first_parent_revwalk(commit.id())
    .map(|rw| IssueMessagesIter {
        inner: rw,
        repo: repo
    })
```

(Although I'm not sure whether the compiler is smart enough to figure this out...)

lib/src/iter.rs
```diff
@@ -61,36 +60,38 @@ impl<'r> Iterator for HeadRefsToIssuesIter<'r>
...
-
-        next
+        self.inner
+            .next()
+            .map(|item| item
+                .and_then(|id| {
+                    self.fuse_if_initial(id);
+                    self.repo.find_commit(id)
+                })
+                .chain_err(|| EK::CannotGetCommit)
```
matthiasbeyer: We could pass the ID here to the error kind to have it in the error output, but that'd include some ... not that easy transformation to the code here at hand...

neithernut: Look again. The error may not come from the `find_commit()` but also from the `Revwalk`.

PR #101, opened 2017-06-07
Clean imports

neithernut:
In some places, we import a specific data structure and use it in only one or two places. This PR targets those imports. Suggestions are welcome.

### DIFF
--- a/lib/src/iter.rs
+++ b/lib/src/iter.rs
@@ -12,7 +12,7 @@
 //! This module provides various iterators.
 //!
 
-use git2::{self, Repository, References};
+use git2::{self, Repository};
 
 use issue;
 use repository::RepositoryExt;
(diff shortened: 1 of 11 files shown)


neithernut:
So you'd rather have every single little type explicitly imported, resulting in a potentially obnoxiously humongous list of `foo::bar::{this, and, that, as, well, as, those, things}`?

matthiasbeyer:
Yes. That's what it is for.

neithernut:
Well, I have a _very_ different stance on that.

neithernut:
Rebased due to merge conflicts.

neithernut:
@matthiasbeyer please give an ACK/NACK on those changes. I'm willing to omit some of them if you feel that way, but some of them just make sense.

Regarding the changes in the logger module, another possibility would be to import `log::*`, since we are implementing a logger anyway.

matthiasbeyer:
> Regarding the changes in the logger module, another possibility would be to import log::*, since we are implementing a logger anyway.

I'm generally against importing `*`, so I'm fine with this part. I'm also fine with the minifications or imports and the removal of the unused things.

Not with the "Importing modules and using the module::type path" part, as said above.

PR #102, opened 2017-06-07
Make apparent module structure of library more convenient

neithernut:
...by renaming and re-exporting some types.


neithernut:
I'm open to suggestions. For example, I can't decide whether or not we should re-export `libgitdit::messages::trailer::Trailer` as `libgitdit::messages::Trailer`.

matthiasbeyer:
> I can't decide whether or not we should re-export 

I'm generally against reexports and, as far as I know, it is not idomatic in rust to reexport types in a library. The reexport feature is for library-internal reexporting, I'd say. Not for exporting types from one library in another.

On a sidenote: I'm also not reexporting `toml::Value` in my `toml_query` crate - if you want to work with `toml_query` you _always_ work with `toml` - and I see the same for this library, at least to some extend.

neithernut:
> 'm generally against reexports and, as far as I know, it is not idomatic in rust to reexport types in a library.

Weren't you the one who said that it was cleaner to re-export the `RepositoryExt` and `Message` traits rather than have them both in `lib/src/lib.rs`?

matthiasbeyer:
Please re-read:

> The reexport feature is for library-internal reexporting, I'd say. Not for exporting types from one library in another.

neithernut:
In this case I don't get your original argument: re-exporting `libgitdit::messages::trailer::Trailer` as `libgitdit::messages::Trailer` is a library-internal re-export in this context.

matthiasbeyer:
Ah, I thought you wanted to export things from `git2`...

PR #103, opened 2017-06-07
Simplify error handling in the binary

neithernut:
We have a lot of ugly stuff in the binary, just for the sake of error handling. This PR aims to make this less painful.

### DIFF
--- a/src/abort.rs
+++ b/src/abort.rs
@@ -7,7 +7,7 @@
 //   published by the Free Software Foundation.
 //
 
-use std::fmt::Debug;
+use logger::LoggableError;
 use std::process::exit;
 
 /// Aborting iterator
@@ -31,7 +31,7 @@ impl<I, V, E> From<I> for AbortingIter<I, V, E>
 
 impl<I, V, E> Iterator for AbortingIter<I, V, E>
     where I: Iterator<Item = Result<V, E>> + Sized,
-          E: Debug
+          E: LoggableError
 {
     type Item = V;
 
@@ -61,7 +61,7 @@ impl<I, V, E> IteratorExt<I, V, E> for I
 
 impl<I, V, IE, OE> IteratorExt<I, V, IE> for Result<I, OE>
     where I: Iterator<Item = Result<V, IE>> + Sized,
-          OE: Debug
+          OE: LoggableError
 {
     fn abort_on_err(self) -> AbortingIter<I, V, IE> {
         AbortingIter::from(self.unwrap_or_abort())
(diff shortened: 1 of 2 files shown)


neithernut:
The idea is to make use of the new extensions in the subcommand implementations. There, it is absolutely valid to simply abort in most cases. Maybe we'll eliminate `try_or_1!` in favour for `unwrap_or_abort()`. The latter is obviously better for chaining stuff.

src/abort.rs
```diff
@@ -64,3 +59,35 @@ impl<I, V, E> IteratorExt<I, V, E> for I
...
+{
+    /// Just like a regulat unwrap() except it performs proper logging
+    ///
+    /// Returns the contained value or aborts the program, logging the error.
+    ///
+    fn unwrap_or_abort(self) -> V;
+}
+
+impl<V, E> Abortable<V> for Result<V, E>
+    where E: Debug
```
matthiasbeyer: I'm not that sure here, but as we're using `error_chain` I'd rather use the actual error type and `error_chain`s tracing functionality and not the `error!(...)` logging macro...

neithernut: The `ErrorChain` trait implements `Display`. The documentation states that the entire chain will be printed in such cases. I assumed that `error!()` will do the right thing in such cases.

neithernut: In any case, I'd keep this implementation if possible and supplement a specialization. We also have errors we get straight from `git2`.

neithernut: Mh... `rust` complains about conflicting implementations.

neithernut:
I'll rebase this branch on the latest master and make use of the new traits. After I'm done we can evaluate how much pain it would be to only support `ChainedError`s with the new type.

neithernut:
I only applied changes which introduced some benefit imo so far, e.g. I didn't eliminate `try_or_1!` completely.

neithernut:
Looking at subcommand implementations, maybe it makes sense to eliminate `try_or_1!` completely in favour for `unwrap_or_abort()` and have the subcommands return "void" instead of an "i32". It would be much cleaner imo. @matthiasbeyer?

matthiasbeyer:
> Looking at subcommand implementations, maybe it makes sense to eliminate `try_or_1!` completely in favour for `unwrap_or_abort()` and have the subcommands return "void" instead of an "i32". It would be much cleaner imo.

Yup.

neithernut:
Most commit messages will be rewritten in the squash

neithernut:
Blocks #104.

neithernut:
Blocks #105.

neithernut:
@matthiasbeyer Are you going to review this anytime soon?

src/abort.rs
```diff
@@ -64,3 +59,35 @@ impl<I, V, E> IteratorExt<I, V, E> for I
...
+        AbortingIter::from(self.unwrap_or_abort())
+    }
+}
+
+
+/// Extension trait for convenient abortion in case of errors
+///
+pub trait Abortable<V>
+{
+    /// Just like a regulat unwrap() except it performs proper logging
```
matthiasbeyer: typo `regulat`

PR #104, opened 2017-06-08
Rethink message creation

neithernut:
`repository::RepositoryExt::create_message()` is a function doing to some extent two different things, dependent on the `issue` parameter. That parameter still is an `Oid` despite the new `Issue` type (well, I simply forgot about it). In this PR, we rethink the `create_message()` interface.

I'd like to replace `create_message()` completely. However, I'm not yet sure with what to replace it. Suggestions are welcome.

### DIFF
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -12,7 +12,7 @@
 //! This module provides the `Issue` type and related functionality.
 //!
 
-use git2::{self, Oid, Reference, References};
+use git2::{self, Commit, Oid, Reference, References};
 use std::fmt;
 use std::result::Result as RResult;
 
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -142,6 +142,18 @@ impl<'r> Issue<'r> {
             .chain_err(|| EK::CannotSetReference(refname))
     }
 
+    /// Add a new leaf reference associated with the issue
+    ///
+    /// Creates a new leaf reference for the message provided in the issue.
+    ///
+    pub fn add_leaf(&self, message: Oid) -> Result<Reference> {
+        let refname = format!("refs/dit/{}/leaves/{}", self.ref_part(), message);
+        let reflogmsg = format!("git-dit: new leaf for {}: {}", self, message);
+        self.repo
+            .reference(&refname, message, false, &reflogmsg)
+            .chain_err(|| EK::CannotSetReference(refname))
+    }
+
     /// Get reference part for this issue
     ///
     /// The references associated with an issue reside in paths specific to the
--- a/lib/src/issue.rs
+++ b/lib/src/issue.rs
@@ -125,6 +125,23 @@ impl<'r> Issue<'r> {
             .chain_err(|| EK::CannotGetReferences(glob))
     }
 
+    /// Update the local head reference of the issue
+    ///
+    /// Updates the local head reference of the issue to the provided message.
+    ///
+    /// # Warnings
+    ///
+    /// The function will update the reference even if it would not be an
+    /// fast-forward update.
+    ///
+    pub fn update_head(&self, message: Oid) -> Result<Reference> {
+        let refname = format!("refs/dit/{}/head", self.ref_part());
+        let reflogmsg = format!("git-dit: set head reference of {} to {}", self, message);
+        self.repo
+            .reference(&refname, message, true, &reflogmsg)
+            .chain_err(|| EK::CannotSetReference(refname))
+    }
+
     /// Get reference part for this issue
     ///
     /// The references associated with an issue reside in paths specific to the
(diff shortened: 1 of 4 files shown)


neithernut:
One option would be to have users use `git2::Repository::commit()` directly and subsequently use `update_head()` and `add_leaf()` in order to create the refs necessary. In the case of new issues, this would require creating a new `Issue` instance from the message/commit id is first.

Another possibility would be to add a function `new_issue()` to `RepositoryExt` which performs the commit, creates the issue and "updates" the head ref. Optionally, we may also add a function `add_message()` to `Issue`, which does the same for leaves.
In any case, the `create-message` subcommand implementation would be a bit more complicated, but it would surely pay off for `new` and `reply`.

lib/src/issue.rs
```diff
@@ -119,6 +125,60 @@ impl<'r> Issue<'r> {
             .chain_err(|| EK::CannotGetReferences(glob))
     }
 
+    /// Add a new mesage to the issue
```
matthiasbeyer: Typo.

lib/src/issue.rs
```diff
@@ -119,6 +125,60 @@ impl<'r> Issue<'r> {
...
+    ///
+    /// Adds a new messaage to the issue. Also create a leaf reference for the
```
matthiasbeyer: Typo.

lib/src/issue.rs
```diff
@@ -119,6 +125,60 @@ impl<'r> Issue<'r> {
...
+    {
+        let parent_vec : Vec<&Commit> = parents.into_iter().collect();
+
+        self.repo
+            .commit(None, author, committer, message.as_ref(), tree, &parent_vec)
+            .and_then(|id| self.repo.find_commit(id))
+            .chain_err(|| EK::CannotCreateMessage)
+            .and_then(|message| {
+                self.add_leaf(message.id())?;
+                Ok(message)
```
matthiasbeyer: I'd rather use the following pattern:

```
.and_then(|msg| self.add_leaf(msg.id()))
.map(|_| message)
```

lib/src/issue.rs
```diff
@@ -119,6 +125,60 @@ impl<'r> Issue<'r> {
...
+            })
+    }
+
+    /// Update the local head reference of the issue
+    ///
+    /// Updates the local head reference of the issue to the provided message.
+    /// The `force` parameter may be set to `true` in order to allow forced
+    /// updates.
+    ///
```
matthiasbeyer: We should note that we do not check whether this is a fast-forward-update!

lib/src/repository.rs
```diff
@@ -60,22 +60,18 @@ pub trait RepositoryExt {
...
-                      issue: Option<&Oid>,
-                      author: &Signature,
-                      committer: &Signature,
-                      message: &str,
-                      tree: &Tree,
-                      parents: &[&Commit]
-                     ) -> Result<Oid>;
+    /// Create a new issue with an initial message
+    ///
+    fn issue<'a, A, I, J>(&self,
```
matthiasbeyer: About the naming here: I'd suggest `create_issue` or something as function name, not only `issue` where one _has_ to look up the documentation on what it does. `create_issue` is a more speaking name, IMO.

lib/src/issue.rs
neithernut: This is not possible, since `add_leaf` doesn't return the message but a (git) reference.

lib/src/issue.rs
neithernut: Isn't that exactly the point of the `force` parameter?

lib/src/issue.rs
matthiasbeyer: Yes, but that would be dropped in the following `map` - or am I missing something in the diff?

lib/src/issue.rs
matthiasbeyer: Yes, but when `force = false` we do not check whether this is actually a `ff`-update - AFAICS.

lib/src/issue.rs
neithernut: If you have just the `add_leaf` invocation in the `and_then` closure, the message is gone.

matthiasbeyer: Ah, then we could use `.and_then(|msg| self.add_leaf(message.id()).map(|_| message))` here!

lib/src/issue.rs
neithernut: Mh.... `reference()` calls `git2_reference_create()` from libgit2. The documentation of that function states that the `force` parameter controls whether an existing ref is updated or not. If the reference exists, it is not fast-forwarded if `force=false`.

The thing is that this function doesn't make sense that way. I consider a modification.

matthiasbeyer: I would update the comment, so the warning is in a section `# Warnings`. Its common for Rust code documentation to do it this way.

matthiasbeyer: Shall we provide a checked version?

neithernut: Yep, but not right now. It's kinda complex, because you have to check whether it's fast-forwardable manually.

matthiasbeyer: Oh okay.

PR #114, opened 2017-07-07
Add various tests for the library

### LINKED ISSUE #78: Add tests
Up until now, we neglected automated testing. All we did was manually smoke-testing new functionality.
To reduce the probability of errors leaking through, we should definitely make use of automated testing, including tests using on-the-fly repos.
Obviously, the library and library-internal modules should be the primary target for testing. However, it was suggested having some smoke-tests on the binary, in order to prevent unintentionally breaking the user interface.

neithernut:
Resolves #78.

### DIFF
--- a/lib/src/message/mod.rs
+++ b/lib/src/message/mod.rs
@@ -202,3 +202,38 @@ impl<'c> Message for Commit<'c> {
     }
 }
 
+
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // LineIteratorExt tests
+
+    #[test]
+    fn empty_message_format_check() {
+        let vec : Vec<&str> = Vec::new();
+        assert!(vec.into_iter().check_message_format().is_err());
+    }
+
+    #[test]
+    fn empty_message_format_check2() {
+        assert!(vec![""].into_iter().check_message_format().is_err());
+    }
+
+    #[test]
+    fn oneline_message_format_check() {
+        vec!["Foo bar"].into_iter().check_message_format().unwrap();
+    }
+
+    #[test]
+    fn malformed_message_format_check() {
+        assert!(vec!["Foo bar", "Baz"].into_iter().check_message_format().is_err());
+    }
+
+    #[test]
+    fn multiline_message_format_check() {
+        vec!["Foo bar", "", "Baz"].into_iter().check_message_format().unwrap();
+    }
+}
--- a/lib/src/test_utils.rs
+++ b/lib/src/test_utils.rs
@@ -1,0 +1,56 @@
+// git-dit - the distributed issue tracker for git
+// Copyright (C) 2017 Matthias Beyer <mail@beyermatthias.de>
+// Copyright (C) 2017 Julian Ganz <neither@nut.email>
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+
+//! Testing utils
+//!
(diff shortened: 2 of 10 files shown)


neithernut:
Meh. The CI fails for the old version because of language features which were unstable back then. As far as I can see, we have the following options:
 * Allow unstable features for old versions
 * Drop support for old versions
 * Disable tests for old versions

The first one would make sense, since we or libs we depend on might end up using now-stable features which were unstable before `rust-1.17.0`. However, at some point, we will drop support anyway as new versions of rust are released. Disabling tests doesn't make much sense, IMO.

Thoughts, @matthiasbeyer?

neithernut:
Note: we'll also have to make Travis perform tests for `libgitdit` rather than for `git-dit` only.

neithernut:
The `Trailers` iterator misbehaves. Apparently, the `match` clause in line 255 of `trailer.rs` does never encounter a plain line of text. I have yet to find out why.

neithernut:
Note: I'll extract fixed from this PR into a separate one at some point.

neithernut:
Rebased due to CI fix.

lib/src/message/mod.rs
```diff
@@ -202,3 +202,41 @@ impl<'c> Message for Commit<'c> {
...
+    #[test]
+    fn oneline_message_format_check() {
+        vec!["Foo bar"].into_iter().check_message_format().unwrap();
+    }
+
+    #[test]
+    #[should_panic]
+    fn malformed_message_format_check() {
+        vec!["Foo bar", "Baz"].into_iter().check_message_format().unwrap();
+    }
```
matthiasbeyer: Why are you not using `assert!()` in the tests? If things do not panic (in the above tests) they might still not return what we would expect. Especially in this test case, if `check_message_format()` panics, we get a green test for this - which is clearly not valid!

```
assert!(vec[..].into_iter().check_message_format().is_err())
```

would be the way to go here.

neithernut: Good point.

neithernut:
The `TestingRepo` type I just introduced may be rebuild on basis of an in-memory key-value store once alexcrichton/git2-rs#222 is resolved.

lib/src/test_utils.rs
```diff
@@ -0,0 +1,56 @@
...
+        ).expect("Could not open/init repository");
+        TestingRepo { repo: repo }
+    }
+
+    /// Get a reference of the repo
+    ///
+    pub fn repo(&mut self) -> &mut Repository {
+        &mut self.repo
+    }
+}
```
matthiasbeyer: What about implementing `Drop` to remove the repository after the test?

You could even make the removal dependent on an environment variable if you want to keep the repository after the test for debugging purposes.

neithernut: I originally planned implementing `Drop` then introducing `TestingRepo`. I didn't do it because I hoped it would be unnecessary to remove directories.

Well, keeping testing repos around has the advantage that it can be examined after a failure. If we implement the purging functionality in `Drop` and let users disable it for a single run, we may still have the repositories around in the next run, resulting in some tests failing.

Btw: if we migrate the testing repo to a key-value storage, we should implement `Drop` so that all objects are dumped to either `stdout` or some file, so we end up with something disectable after failed tests.

neithernut:
Note: I originally intended extracting fixes resulting from this test-writing campaign. However, only a single bug surfaced and I think I'll keep the fix in this branch.

neithernut:
@matthiasbeyer: could I get an ACK/comments?

matthiasbeyer:
Not earlier than after my last exam + after the Party (so on friday, not earlier).

neithernut:
Blocks #120 due to CI stuff.

PR #113, opened 2017-06-28
Improvements of the documentation

neithernut:
Resolves #111.

PR #115, opened 2017-07-07
Don't make the CI build for old rust-vresions

neithernut:
Apparently, some dependencies don't care about old rust-versions. Hence,
supporting those old versions doesn't make sense.


neithernut:
Blocks #113, #114.

PR #119, opened 2017-07-13
Metadata accumulation

neithernut:


neithernut:
Regarding the `ValueAccumulator`: another design option would be to have a trait instead of an enum and an implementation for each type of accumulator. With such a setup we wouldn't need to fiddle around with an iterator for retrieving a value if we already know that, in fact, only a single value may be present after accumulation. However, we would also need to either wrap those in an enum or boxes for the accumulate-values-in-a-map use-case -- explicitly, since rust doesn't know "variadic generics".

PR #120, opened 2017-07-13
Make the man page a feature of the Cargo package

neithernut:


neithernut:
Maybe this belongs into the `0.2.1` release. However, it changes the way the package is built in a breaking way. Suggestions @matthiasbeyer?

matthiasbeyer:
I really don't know. I would just wait and include it in a 0.3.0.. no need to push a 0.2.1 for this tiny thing, IMO.

neithernut:
Agreed. We planned on pushing a `0.2.1` release for the tests and the doc though.

PR #124, opened 2017-07-19
Determination of dit reference type

neithernut:
Authors of git server hooks (e.g. `update` hooks) may want to determine
of what type a given dit reference and whether it is a dit reference at
all.


neithernut:
Maybe we should also require that ids in reference names are full hashes.

PR #127, opened 2017-07-24
Categorize blocks of lines rather than single lines

neithernut:
We currently offer line categorization through the `message::line` module and associated types. However, the categorization of single lines is, by definition, not context-aware: a line appearing to be a trailer will be categorized as a trailer even if it is surrounded by regular lines of text. We thus introduced a workaround in the `Trailers` type, where we require correct context-aware categorization of trailers.

This patch-set aims at resolving this design-flaw by replacing the aforementioned `message::line` module with a `message::block` module. The new module will offer categorization of blocks rather than of single lines.


neithernut:
The `Trailers` test currently fails. I have yet to fix this.

PR #128, opened 2017-07-25
Metadata specification

neithernut:
We currently offer both tools for writing and accumulating metadata in the library. However, we lack some specification of metadata which we expect for issues.


neithernut:
I'll probably add some convenience conversation functionality for the different `Accumulator`s.

PR #117, opened 2017-07-12
Collection of old references

neithernut:
This PR introduces garbage collection facilities.

PR #129, opened 2017-07-25
Open the repository in the individual subcommands

neithernut:
We currently open the repository in the main function. This has the serious drawback that `git-dit` will look for a repository and abort on failure even if the subcommand invoked does not require a repository. Such subcommands include "check-message".


neithernut:
This will generate a merge-conflict with #117 and maybe others.

neithernut:
Will generate a merge-conflict with #132.

PR #131, opened 2017-08-03
Consolidate message iteration

neithernut:
Currently, we offer multiple iterators for iterating over messages. Sadly, most of those are isolated from each other. Additionally, we often provide only a `Revwalk` for general purpose usage when the goal is to iterate over messages.


neithernut:
Blocks #132.

PR #133, opened 2017-08-06
Add an example update hook for realizing a dit-only policy

neithernut:
Server administrators may use the hook or parts of for dit-repositories
on their servers.

### DIFF
--- a/examples/hooks/update
+++ b/examples/hooks/update
@@ -1,0 +1,83 @@
+#!/bin/bash
+#
+# Copyright (C) 2017 Matthias Beyer <mail@beyermatthias.de>
+# Copyright (C) 2017 Julian Ganz <neither@nut.email>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the “Software”), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


neithernut:
It is currently untested. Nevertheless, I'd appreciate a review by someone with more experience with bash-scripts, e.g. @manuelmessner

Profpatsch:
For basic sanity testing you can use [shellcheck](https://github.com/koalaman/shellcheck)

manuelmessner:
I don't know how to test this (as I'm not part of the project itself).
But I can have a look at the script:

In general you do not need to quote variables within `[[ ... ]]` because word splitting is not performed there. They are needed if the string on the right hand side consists of special chars which could involve globbing (depending on the bash configuration) or regular expressions and you want to prevent that.
But in your case it should be fine.
Although I would recommend removing them at the `-n` and `-z` tests because there might exist an edge case I cannot think of right now.
And I recommend quoting the `git dit check-refname` call in the `case` statement.

You could probably change the  `echo` within the `die` function to this: `echo "$*" >&2` .
That allows you to write things like `die This is an error message` (as multiple arguments). But that's just an optical change.

neithernut:
Thanks for the review.

neithernut:
The script is now tested and updated accordingly.

PR #143, opened 2017-08-16
Refactor the binary

neithernut:
The binary is a bit messy. This PR tries to lower the level of messiness a bit by moving some modules into new modules, moving the responsibility of aborting into utility functions and cleaning up some old utilities.


neithernut:
Rebased due to conflicts.

PR #145, opened 2017-08-19
Promote "trailer" to a top-level module

neithernut:
Much of the functionality in the `message` module really concerns `Trailer`s rather than `Message`s. In the future, `Trailer` specific functionality may even grow further, for example if the library gains filtering utilities.

This PR promotes the `trailer` module to the top-level. Multiple other modules will be moved from `message` to `trailer` and parts be adjusted. Another goal is to remove dependencies from the `trailer` module to the `message` module (but only the other way).

Probably conflicts with #143.


neithernut:
This is hopefully the last big breaking interface change of the library.

PR #148, opened 2017-08-25
Add issue filtering as a library feature

neithernut:
Filtering issues based on metadata should be a relatively common task in applications associated with git-dit, including web-viewers and of course the `git-dit` binary. It therefore makes sense to provide some functionality associated with filtering in the library.

### DIFF
--- a/lib/src/trailer/accumulation.rs
+++ b/lib/src/trailer/accumulation.rs
@@ -103,6 +103,35 @@ pub trait Accumulator {
     }
 }
 
+
+/// Trait for accumulators accumulating multiple values
+///
+/// # Note
+///
+/// This trait really is a convenience trait for consolidating mapping
+/// containers. It only exists because the standart library doesn't provide
+/// any matching traits (the `Index` trait is not an option).
+///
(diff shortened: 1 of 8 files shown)


lib/src/trailer/accumulation.rs
```diff
@@ -103,10 +104,27 @@ pub trait Accumulator {
...
-//       appropriate map trait in `std`.
-impl<S> Accumulator for collections::HashMap<String, ValueAccumulator, S>
-    where S: BuildHasher
+
+/// Trait for accumulators accumulating multiple values
+///
+/// # Note
+///
+/// This trait really is a convenience trait for consolidating mapping
+/// containers. It olny exists because the standart library doesn't provide
```
matthiasbeyer: typo

neithernut:
Its again type to revert to cbedee917dbe4ce89e45b82a43670df7635e92a6 and take another approach for the filter realization in the binary.

neithernut:
Squashed for merge. I'll leave this lying around for a bit, though.

PR #150, opened 2017-08-28
Flexible message formatting

neithernut:
Currently, the formatting of messages is hard-coded, for example for the "show" subcommand. This is not flexible regarding configurability via command line options. In particular, it does not allow a user defined formatting.

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -54,6 +54,7 @@ mod utils;
 mod test_utils;
 
 // A selection of types are reexported for more convenient access.
+pub use error::Error;
 pub use issue::Issue;
 pub use message::Message;
 pub use remote::RemoteExt;
--- a/src/display/formatter.rs
+++ b/src/display/formatter.rs
@@ -195,3 +195,30 @@ impl<I, J, T, K, B> Iterator for FormattedLines<I, J, T, K, B>
     }
 }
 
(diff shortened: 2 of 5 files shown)


neithernut:
The generic part is mostly done. The only thing I plan to add at this point is more documentation.

The specific formatter for messages will probably be written during the week.

neithernut:
Mh... I have no idea why compilation of the offending seciotn fails all of a sudden.

neithernut:
Ah... there was a `git2` update. Maybe it introduced some breaking changes. That would also explain why I did not encounter the problem here.

I can't build right now but I'll update the dependencies and see whether it will reproduce or not.

neithernut:
More workarounds for compiler restrictions which still appear somewhat artificial to me.

Anyway, the current compilation failure is indeed caused by the `git2` update. Sadly, the fix is not as obvious as expected.

neithernut:
Maybe it would be a good idea to have a macro for assembling formatters. It would take a list of tokens, add an `.into()` to each one and wrap the whole thing in `vec![]`. Thoughts, @matthiasbeyer?

neithernut:
Note: we are using message formatting tokens for issues for now. At some point, we will probably introduce issue formatting tokens featuring issue specific things.

PR #151, opened 2017-08-30
Pre-collect names of local References in push_impl

neithernut:
With git2-0.6.7, the signature of the `References::names` changed (see alexcrichton/git2-rs#245). The function now takes a mutable reference and imposes an additional lifetime constraint on the returned iterator. The latter can not out-life the `References` struct any more. This lead to compilation failures.

The problem is resolved by pre-collecting all names as `String`s within a closure.


neithernut:
Blocks virtually everything.

PR #152, opened 2017-08-31
Implement line writing on line iterators rather than streams

neithernut:
We often create line iterators for writing via multiple steps. Currently, those iterators are bound to a name for readability, just to be passed to `WriteExt::consume_lines()`.

Implementing a `write_lines()` function on line iterators instead improves usability in conjunction with chaining, since the stream is usually retrieved more easily.


neithernut:
Conflicts with #150.

PR #154, opened 2017-09-01
Enable alternative author and date via cli parameters

neithernut:
git commit` allows setting the author and date of a commit via command line parameters. This PR introduces a similar interface for the "create-message", "new" and "reply" subcommands. The new interface may be used for scripting, e.g. for implementing a mail-to-message bridge.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -27,6 +27,11 @@ error_chain! {
             display("Malformed author: {}", author)
         }
 
+        MalformedDate(date: String) {
+            description("Malformed date (expected rfc3339/iso8601)")
+            display("Malformed date: {}", date)
+        }
+
         UnknownMetadataKey(key: String) {
             description("Unknown metadata key")
             display("Unknown metadata key: {}", key)
--- a/src/error.rs
+++ b/src/error.rs
@@ -22,6 +22,11 @@ error_chain! {
             display("Malformed filter spec: {}", spec)
         }
 
+        MalformedAuthor(author: String) {
+            description("Malformed author (expected author formatted like 'Foo Bar <foo@bar.net>')")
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -22,6 +22,11 @@ error_chain! {
             display("Malformed filter spec: {}", spec)
         }
 
+        MalformedAuthor(author: String) {
+            description("Malformed author")
```
matthiasbeyer: We should maybe state how the format should be.

src/util.rs
```diff
@@ -179,6 +184,55 @@ impl<'r> RepositoryUtil<'r> for Repository {
...
 
+    fn cli_author(&self, matches: &ArgMatches) -> git2::Signature {
+        use chrono::{DateTime, FixedOffset};
+
+        // First, get a signature appropriate for the author
+        let author = matches
+            .value_of("author")
+            .map(|a| {
+                // NOTE: this function will be called only once per run anyway
+                let parts = Regex::new(r"^([^<>]+)[[:space:]]+<([^[:space:]@]+@[^[:space:]@]+)>$")
```
matthiasbeyer: This regex could be put either in the outer scope or even as constant into a `lazy_static`, so the regex is only compiled once per call of `cli_author()` or even only once per `git-dit` execution!

neithernut: Have you even read the note one line above? If this function is called more than once per run, the dev using it is doing something wrong.

neithernut:
I think that's purely a fault of the documentation. The code does override the author date, not the committer date.

src/util.rs
matthiasbeyer: Whops, jup.

PR #158, opened 2017-09-14
Preparations for the 0.4.0 release

neithernut:
The release is scheduled for September the 15th.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,67 @@
 # git-dit ChangeLog
 
+## v0.4.0 (2017-09-15)
+
+### Binary
+
+Changes:
+ * The message formatting for the "show" subcommand was improved.
+ * In filter-specs for the "list" subcommand, a '=' is used for separating key
+   and value when filtering for an exact value, rather than a ':'.
+
+Added features:
+ * The "create-message", "new", "reply" and "tag" now accept the "--author" and
+   "--date" options. Analogue to the options of "git commit" with the same name,
+   the new options alter the author and authoring date of the resulting message.
+ * New "reporter-name" and "reporter-email" filter keys for issue filtering
+   using the "list" subcommand. As the name suggests, the keys may be used for
+   filtering issues based on the reporter's name or email.
+ * Filter specs for the "list" subcommand can now be used for filtering by
+   substring or presence of a piece of issue metadata.
+ * Filter specs for the "list" subcommand can now be inverted.
+
+Bug-fixes:
+ * Git-dit did not handle pager and editor flags correctly. For example,
+   configuring `less -R` as the pager (contrary to `less`) would cause git-dit
+   search for a binary named "less -R", resulting in an error.
+
+### Library
+
+Changes:
+ * `LineIteratorExt::stripped()` now returns an iterator which also strips
+   trailing blank lines.
+ * `Issue::new()` now returns a `Result`.
+ * Multi-line trailers parsed by the `Blocks` iterator are represented as String
+   `TrailerValue` with newlines separating the individual lines from the input.
+ * `trailer` was promoted to a top-level module.
+ * `message::accumulation` was moved to `trailer`
+ * `message::metadata` was moved to `trailer` and renamed to `spec`.
+ * `MetadataSpecification` was renames to `TrailerSpec`.
+ * `PairsToTrailers` and `DitTrailers` iterator was moved to new
+   `trailers::iter` module.
+ * `Trailers` iterator was moved to `message::block` module.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,67 @@
...
+ * In filter-specs for the "list" subcommand, a '=' is used for separating key
+   and value when filtering for an exact value, rather than a ':'.
+
+Added features:
+ * The "create-message", "new", "reply" and "tag" now accept the "--author" and
+   "--date" options. Analogue to the options of "git commit" with the same name,
+   the new options alter the author and authoring date of the resulting message.
+ * New "reporter-name" and "reporter-email" filter keys for issue filtering
+   using the "list" subcommand. As the name suggests, the keys may be used for
+   filtering issues based on the reporter's name or email.
```
matthiasbeyer: Unclear what a "reporter" is in this context: Is it the author of the issue (as in the git-commit author) or a dit-trailer `Dit-author:` or something like that. Please clearify.

CHANGELOG.md
```diff
@@ -1,5 +1,67 @@
...
+ * Filter specs for the "list" subcommand can now be inverted.
+
+Bug-fixes:
+ * Git-dit did not handle pager and editor flags correctly. For example,
+   configuring `less -R` as the pager (contrary to `less`) would cause git-dit
+   search for a binary named "less -R", resulting in an error.
+
+### Library
+
+Changes:
```
matthiasbeyer: Are these changes backwards-compatible? As I think they are mostly not, I would state that. Of course we are pre-1.0, but it doesn't hurt to say "Backwards-incompatible Changes" here rather than not specifying exactly what this list is about.

matthiasbeyer:
I think we should also add the `Cargo.lock` to the repository for the 0.4.0 release.

matthiasbeyer:
Another note: The `README` lists the dependencies, but the list is incomplete.

matthiasbeyer:
Aaaand of course the binary is not yet published on crates.io - not the 0.2.0 and not the 0.3.0 release. I guess its time, we should put 0.4.0 up there.

CHANGELOG.md
neithernut: I consider points in "Changes" sections generally backwords-incompatible. And we used the term in past releases.

neithernut:
> I think we should also add the `Cargo.lock` to the repository for the 0.4.0 release.

Just nope.

> Aaaand of course the binary is not yet published on crates.io - not the 0.2.0 and not the 0.3.0 release. I guess its time, we should put 0.4.0 up there.

I thought I was clear on this topic: I will not publish the binary on https://crates.io.

CHANGELOG.md
neithernut: This information does not belong in the Changelog but in the manpage/user documentaton.

Because of the "list" subcommand alone, we should introduce subcommand-specific manpages imo, btw.

matthiasbeyer:
> Just nope.

Without it, packaging becomes utterly complex.

> I thought I was clear on this topic: I will not publish the binary on https://crates.io.

Then I will.

neithernut:
> Without it, packaging becomes utterly complex.

Packagers can (and will) ship a `Cargo.lock` which exactly fits their needs. If we ship one, they will most probably patch it anyway. Also, I will not pollute the repository because of the shortcomings of cargo.

matthiasbeyer:
Whatever. I just know that I am not able to package `git-dit-0.3.0` for nixpkgs _because of the missing Cargo.lock file_. In fact, I'm not even able build 0.3.0 anymore because of the missing _Cargo.lock_. Its not like user experience matters at all... so whatever.

CHANGELOG.md
matthiasbeyer: I'm against subcommand specific manpages,... having one huge manpage is better for user experience because you can simply search through one manpage and you know that if you did not find what you were looking for, it is not there.

At least with the current amount of features, multiple manpages make no sense IMO. Of course we could reconsider this if we hit a 1.0 for example, and there are much more flags and subcommands around...

TheNeikos:
Since you have a library and a executable in the same repository it makes sense to distinguish between the two when it comes to the `Cargo.lock`.

http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

Check here for more information

vyp:
A Cargo.lock is recommended to be checked in to version control for executable projects: http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

I personally wouldn't consider it pollution since it actually helps make the build more reproducible.

Issue #159, opened 2017-09-30
Neither --verbose nor --trace print any output


matthiasbeyer:
At least not with `push`.

matthiasbeyer:
`--debug` does also not print any output.

PR #162, opened 2018-01-16
Update dependencies

matthiasbeyer:
All changes tested with `cargo check`, the `log` update tested by executing (after all patches applied).

Because of #159 "log" might not be properly tested.


matthiasbeyer:
The lib actually does not need `log`.

matthiasbeyer:
Ping.

neithernut:
Could you rebase this branch?

matthiasbeyer:
Rebased and also updated `git2` to `0.7`. Lets see whether this still compiles with the new version of `git2`.

neithernut:
Nope, but the issue may be resolved easily.

matthiasbeyer:
I'm having a meetup now. We need to pass [this](https://docs.rs/git2/0.7.1/git2/struct.Sort.html) instead of the constants from 0.6, which vanished in 0.7.

neithernut:
I already prepared some fixes. You should get them soon.

PR #163, opened 2018-01-16
Colorize output

matthiasbeyer:
Simple proposal for colored output.

No flags for disabling or changing of color yet. Might be an idea to add, though I guess this first proposal might suffice.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ is-match = "0.1"
 lazy_static = "0.2"
 log = "0.3"
 regex = "0.2"
+colored = "1.6"
 
 [dependencies.clap]
 version = "2.23"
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Ping.

neithernut:
Colouring should have a low priority, e.g. compared to internationalization. Hence I'd opt for postponing this one until after internationalization is introduced (both will affect the output)

matthiasbeyer:
The problem is that there is no internationalization mechanism in rust, yet. Hence we can consider internationalization a long-time effort rathern than solvable in a short time period.

Thus I'd rather merge this sooner than later. Adding internationalization will require a refactoring anyways, I assume. So there is no point in keeping this out of the codebase, at least from a planning perspective.

PR #164, opened 2018-01-16
Fix warnings

matthiasbeyer:
Small fixes for removing compiler warnings from the compile output.

The second change might be redone by removing the enum variants, I'm not sure on that one.

### DIFF
--- a/src/display/msgtree.rs
+++ b/src/display/msgtree.rs
@@ -188,7 +188,7 @@ impl<'r, I> Iterator for TreeGraphElemLineIterator<'r, I>
             let mut parent_update = commit.parent(0).as_ref().map(Commit::id).ok();
 
             // generate graph elements for the parents currently tracked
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Ping.

src/display/msgtree.rs
```diff
@@ -188,7 +188,7 @@ impl<'r, I> Iterator for TreeGraphElemLineIterator<'r, I>
             let mut parent_update = commit.parent(0).as_ref().map(Commit::id).ok();
 
             // generate graph elements for the parents currently tracked
-            let mut elems : TreeGraphElemLine = self.parents.iter_mut().map(|mut parent| {
+            let mut elems : TreeGraphElemLine = self.parents.iter_mut().map(|parent| {
```
neithernut: Not sure about this one. It might be one of those cases where people didn't consider the implications of semantic versioning, e.g. it is mandatory if a specific version of $whatever is used.

Issue #161, opened 2018-01-16
Flag to _not_ use pager


matthiasbeyer:
A flag for _not_ piping to pager is missing.

Also: If the terminal is wide/high enough, we shouldn't pipe to the pager at all (or provide a flag to enforce it).

matthiasbeyer:
Also, if `stdout` is not a tty, we shouldn't call the pager. Things like `git dit list | head` do not work.

neithernut:
This is already handled by `git`. E.g. run `git --no-pager dit list`.

PR #165, opened 2018-01-31
Do not call pager if we are printing to a pipe

matthiasbeyer:
Rel #161

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ is-match = "0.1"
 lazy_static = "0.2"
 log = "0.3"
 regex = "0.2"
+atty = "0.2"
 
 [dependencies.clap]
 version = "2.23"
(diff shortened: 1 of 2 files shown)


neithernut:
Did you actually test whether git doesn't already do the right thing here?

matthiasbeyer:
Yes, I tested it and it broke. But funnily enough, I can not reproduce this issue right now (neither with `git dit` and the `git-dit` binary in `$PATH` nor with calling `git-dit` directly). Would not merge until further investigation.

neithernut:
I just did a quick test with a shell-script which I named `git-printenv` in order to make it a subcommand:

```
#!/bin/sh

echo $GIT_EDITOR
echo $GIT_PAGER
git config core.editor
git config core.pager
echo $EDITOR
echo $PAGER
```

It appears that `git` does indeed not alter the environment variables or the configuration for non-interactive shells (tested by piping the output to `cat`). Hence, we should detect non-interactive shells. However, I would not do this at the call-site but rather handle it inside `programs.rs`. E.g. return a handle for accepting the input, which is either stdout or a pipe. Or unconditionally try to run `cat`.

neithernut:
The issue was resolved by #180.

PR #167, opened 2018-01-31
Add no-pager for "show" and "list" commands

### LINKED ISSUE #161: Flag to _not_ use pager

matthiasbeyer:
Closes #161 

Will conflict with #165


neithernut:
Nope.

PR #173, opened 2018-03-23
Add nixpkgs package name

wmertens:


neithernut:
Normally, I'd not put information about packaged versions in the README but on a project website. I rather consider this another bad habit people get tought on GitHub. However, we currently lack a proper website and I'm not willing to put one up just yet.

So for the time being I'll merge this PR. But I'll revert it.

Issue #174, opened 2018-03-26
git-dit gc is slow


matthiasbeyer:
Hi,

as I already told you in private, I am playing with importing all issues from [imag](https://github.com/matthiasbeyer/imag) into git-dit.

I created [git-dit-github-import-rb](https://github.com/matthiasbeyer/git-dit-github-import-rb/) and imported the issues. Now I can `git dit gc --dry-run`. It took _way_ too long.

I have about 8500 commits created by the import, `git dit gc --dry-run | wc -l` prints almost 1900 refs which will be garbage collected, but it takes 51 seconds (in debug build) to complete the dry-run, which is too long IMO.

Also it blocks until all refs are found and prints them after the aggregation. I don't know how much that might be improvable, but we should consider printing as early as possible here, IMO.
Also, some status output would be really nice, for example using [indicatif](https://crates.io/crates/indicatif), which is a really nice crate for progress reporting IMO.

Maybe we can add parallelization in the garbage collector implementation to speed things up a good bit.

neithernut:
IIRC, the garbage refs are collected because of a mixture of reducing code duplication and another `Result<Iter<Result>>`. The collection code itself should return an iterator which may be used directly. 

However, this does not change a thing about the GC being slow. I'll try to have a look soonish.

neithernut:
Note: while making the GC code capable of doing things in parallel might be of value in the future, I'd rather figure out why it's slow in the first place, because it shouldn't be.

neithernut:
Maybe we are not to blame, after all: libgit2/libgit2#4428
Note: this is just a speculation, I'll make proper measurements some time in the future.

neithernut:
@matthiasbeyer do you have a repo somewhere for reproduction/measurement?

neithernut:
I obtained a repo for testing from @matthiasbeyer. I can somewhat reproduce the (low) performance observed by him:
```
$ time git-dit gc --dry-run
real	0m29.135s
user	0m10.965s
sys	0m17.705s
```

According to `perf`, nearly 80% of the time is spent in `git2::repo::Repository::references_glob` or `git_reference_iterator_glob_new` (the latter is the `libgit2` function invoked by the `git2` abstraction).

[...]

PR #178, opened 2018-04-23
Refactor the GC code

neithernut:
This is somewhat related to #174.

This patch-set decouples the collection of references for different issues. Leaves referenced by refs of another issue will no longer be collected. Issues are no longer supplied up-front via `CollectableRefs::new()` but are fed into `CollectableRefs::into_refs()` and `CollectableRefs::into_collector()`. However, users are discouraged from using those functions now. Instead, we introduces a new function `CollectableRefs::for_issue()` which allows collecting references for a single issue without mutating or consuming the `CollectableRefs`. It enables more direct updates, e.g. when using a `FlatMap` iterator.

### DIFF
--- a/lib/src/gc.rs
+++ b/lib/src/gc.rs
@@ -193,8 +193,17 @@ impl<'r> CollectableRefs<'r>
         Ok(retval)
     }
 
-    /// Transform directly into a reference collection iterator
+    /// Produce a reference collection iterator for multiple issues
     ///
+    /// This is a convenience function.
+    ///
+    /// # Note
+    ///
+    /// Internally, this function collects the references during the call.
+    /// Consider using the `for_issues` function in conjunction with
+    /// `flat_map()` on an issue iterator instead.
+    ///
+    #[deprecated]
     pub fn into_collector<I, J, K>(self, issues: I) -> Result<ReferenceCollector<'r>>
     where I: IntoIterator<Item = K, IntoIter = J>,
           J: Iterator<Item = K>,
--- a/lib/src/gc.rs
+++ b/lib/src/gc.rs
@@ -45,33 +45,26 @@ pub enum ReferenceCollectionSpec {
 /// Use this type in order to compute dit-references which are no longer
 /// required and thus may be collected.
 ///
-pub struct CollectableRefs<'r, I, J = Issue<'r>>
-    where I: Iterator<Item = J>,
-          J: Borrow<Issue<'r>>
+pub struct CollectableRefs<'r>
 {
     repo: &'r git2::Repository,
-    issues: I,
     /// Should remote references be considered during collection?
     consider_remote_refs: bool,
     /// Under what circumstances should local heads be collected?
     collect_heads: ReferenceCollectionSpec,
 }
 
-impl<'r, I, J> CollectableRefs<'r, I, J>
-    where I: Iterator<Item = J>,
-          J: Borrow<Issue<'r>>
+impl<'r> CollectableRefs<'r>
 {
     /// Create a new CollectableRefs object
     ///
     /// By default only local references are considered, e.g. references which
     /// are unnecessary due to remote references are not reported.
     ///
-    pub fn new<K>(repo: &'r git2::Repository, issues: K) -> Self
-        where K: IntoIterator<Item = J, IntoIter = I>
+    pub fn new(repo: &'r git2::Repository) -> Self
     {
         CollectableRefs {
             repo: repo,
-            issues: issues.into_iter(),
             consider_remote_refs: false,
             collect_heads: ReferenceCollectionSpec::Never,
         }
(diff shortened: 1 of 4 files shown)


neithernut:
Once again, I hit a lifetime-related failure which is probably introduced by some elision which does _not_ turn up in the compiler diagnostics...

neithernut:
Ok, I managed to fix it on my own. _Of course_ it was some elision being performed completely silently, which did not turn up in the diagnostics either.

neithernut:
While working on this, I realized something: the GC constructs one common `RefsRefferingTo` iterator for the leaves of all issues fed to the GC. This obviously means that the collection of refs for issues all contribute to a common state, which makes decoupling things harder. We did this because this way leaf references are also collected when they are referenced by messages of _another_ issue.

However, this does not really makes sense. The leave still belongs to the issue and should therefore be found, e.g. when using `git dit show`. Hence, the original design was flawed in this regard. We can fix this, decouple the collection of references for issues and gain the ability to paralellize garbage collection. Hooray!

Downside: I'll stash everything I have done so far (again) and start anew.

lib/src/iter.rs
```diff
@@ -243,6 +252,16 @@ impl<'r> RefsReferringTo<'r> {
...
+
+    /// Preload the iterator with references
+    ///
+    /// The references supplied will be returned by the iterator.
+    ///
+    pub fn preload<I>(&mut self, references: I)
+        where I: IntoIterator<Item = git2::Reference<'r>>
+    {
+        self.current_refs.extend(references);
+    }
```
neithernut: We cold also implement `Extend`, which would spare us a pointless collection within the GC. But I felt it would be wrong, semantically. What's your opinion, @matthiasbeyer?

lib/src/gc.rs
```diff
@@ -98,106 +91,129 @@ impl<'r, I, J> CollectableRefs<'r, I, J>
...
+    ///
+    /// This is a convenience function.
+    ///
+    /// # Note
+    ///
+    /// Internally, this function collects the references during the call.
+    /// Consider using the `for_issues` function in conjunction with
+    /// `flat_map()` on an issue iterator instead.
+    ///
+    pub fn into_refs<I, J, K>(self, issues: I) -> Result<Vec<Reference<'r>>>
```
matthiasbeyer: I don't understand why this does return a `Vec<_>` instead of an custom iterator type.

If this would return a custom iterator type, the `into_collector()` function would also return a custom iterator - which is what it does, and the `ReferenceCollector` type stores an iterator internally. So effectively, we create a `Vec` that we turn into an iterator type again in case of the `into_collector()` function.

In case of the other use of this function, which is in `push_ref_parents()`, we turn the `Vec` directly into an iterator again and do some computation on it, just to collect it afterwards. Which is not very efficient either.

So I guess it would be really beneficial to let this function return an iterator type as well!

lib/src/iter.rs
matthiasbeyer: I think it is rather fitting semantically. I'm not sure what you mean by

>  spare us a pointless collection within the GC

though.

lib/src/gc.rs
```diff
@@ -98,106 +91,129 @@ impl<'r, I, J> CollectableRefs<'r, I, J>
...
-        }
+    /// Find collectable references for multiple issues
+    ///
+    /// This is a convenience function.
+    ///
+    /// # Note
+    ///
+    /// Internally, this function collects the references during the call.
+    /// Consider using the `for_issues` function in conjunction with
+    /// `flat_map()` on an issue iterator instead.
```
matthiasbeyer: Why making this function `pub` if you discourage users here? Why not simply make it private?

lib/src/gc.rs
```diff
@@ -98,106 +91,129 @@ impl<'r, I, J> CollectableRefs<'r, I, J>
...
+    /// This is a convenience function.
     ///
-    pub fn into_collector(self) -> Result<ReferenceCollector<'r>> {
-        self.into_refs()
-            .map(ReferenceCollector::from)
+    /// # Note
+    ///
+    /// Internally, this function collects the references during the call.
+    /// Consider using the `for_issues` function in conjunction with
+    /// `flat_map()` on an issue iterator instead.
```
matthiasbeyer: Same here: Why not making this private?

lib/src/gc.rs
neithernut: The plan is to remove this function, eventually. So I don't see any point in adding yet another custom iterator just for this function.

lib/src/iter.rs
neithernut: https://github.com/neithernut/git-dit/blob/9c039bba0803029f79db532e3276f56603245366/lib/src/gc.rs#L143

lib/src/gc.rs
neithernut: It was `pub` previously. Effectively, it's a legacy API which will vanish at some point.

lib/src/gc.rs
neithernut: Same reason.

lib/src/gc.rs
matthiasbeyer: We could mark it as deprecated then: https://stackoverflow.com/questions/30757914/can-i-mark-a-function-as-deprecated

lib/src/gc.rs
matthiasbeyer: Mid-May the impl-trait feature will be in stable Rust. We can then move forward and return a trait here, I guess.

neithernut: Mh... I'll probably re-iterate over the function, and maybe remove the deprecated-annotation I just applied.

We have a few other places where we would greatly benefit from an impl-trait feature btw.

PR #179, opened 2018-04-30
Enhance the gc subcommand

### LINKED ISSUE #174: git-dit gc is slow

neithernut:
This patch-set introduces multiple enhancements to the `git dit gc` subcommand. The main goal is to enable faster/more "responsive" collection.

It also enables collection for a subset of issues. This functionality may be used in server-side hooks, greatly enhancing the performance.

Resolves #174.

### DIFF
--- a/src/cli.yaml
+++ b/src/cli.yaml
@@ -210,6 +210,13 @@ subcommands:
                 help: Also collect local heads
                 multiple: false
                 takes_value: false # This may change in the future
+            - issue:
+                help: >
+                        Issue for which to collect references (collects for all
+                        issues if not specified)
+                index: 1
+                required: false
+                multiple: true
 
     - list:
         about: >
--- a/src/main.rs
+++ b/src/main.rs
@@ -222,22 +222,30 @@ fn fetch_impl(matches: &clap::ArgMatches) {
 /// gc subcommand implementation
 ///
 fn gc_impl(matches: &clap::ArgMatches) {
-    use libgitdit::gc::{ReferenceCollectionSpec, ReferenceCollector};
+    use libgitdit::gc::ReferenceCollectionSpec;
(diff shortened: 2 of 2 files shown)


neithernut:
I also planned to offload `RefsReferringTo` generation to another thread using [ppipe](https://crates.io/crates/ppipe), but I did not get it to compile, yet. Also, since I have to juggle with `Mutex`es, it may not be worth it. Maybe I should just use a channel, directly.

neithernut:
Turns out offloading work to another thread is not an easy thing to do. Most relevant `git2` types are `!Send`. Wrapping stuff into `Mutex`es doesn't really help, since we (and this includes `git2`-internal code) uses lots of references. According to its docs, libgit2 is thread-safe, but git2-rs doesn't tell the compiler about this (see alexcrichton/git2-rs#194).

So we have to wait until _someone_ does the necessary work of flagging the necessary structs. Until then, it doesn't make much sense trying to implement threading-features at our level.

PR #190, opened 2020-03-12
Update dependency: git2

matthiasbeyer:
This does still result in a lot of warnings during compilation, but at
least it compiles on a current rustc version (1.40.0 for me here).

People want to package this, and an old openssl version holds them back.
openssl is pulled in via the git2 crate, so update it here.


neithernut:
I'll check this out after work, then probably merge right away.

There were some major enhancements in `libgit2` since that old release as well as some usability-enhancements to `git2-rs`. We may be able to eliminate some explicit `collect()`s after the upgrade.

matthiasbeyer:
Maybe, sure. But this is the quick-fix to get the sources to build again and with updated openssl, so I'd rather merge this soon and optimize later.

neithernut:
Side-note: we should decide on a minimum supported (rust) version. I'll probably look into that this weekend.

